PR #2, opened 2021-04-26
Update manifests for new location of Observatorium API

onprem:


jsonnetfile.lock.json
```diff
@@ -109,7 +120,7 @@
           "subdir": "jsonnet/lib"
         }
       },
-      "version": "937be40bc3199cc43ba0bcb2f5c2b8cc5b21cb0f",
+      "version": "0164651421f2d1b77db703bbeea501cb5b7a10cc",
```
squat, line 123: We should remove this dependency, no? Or is this a transitive dep?

onprem: It's a transitive dependency because we haven't updated observatorium/deployments yet. I think it will go away automatically once we update that.

bwplotka (approved):

squat (approved):

PR #9, opened 2021-05-18
Add loki tenant alert to notify on rate limiting events

periklis:
This PR adds another non-oncall-paging alert for Loki when tenants exceed their pre-configured limits. The tenet is to address rate limiting events pro-actively and inform tenants about possible client misconfiguration.


kakkoyun (approved):
LGTM 🚀

PR #10, opened 2021-05-19
Fix prometheus rules annotations

periklis:
The following change set provides minor fixes for our prometheus rules' annotations:
- Dashboard links get a proper value for `var-namespace` based on alert group and environment.
- Add missing dashboard ID for loki tenant alerts
- Fix namespace selector for loki tenant alerts

### DIFF
--- a/observability/prometheusrules.jsonnet
+++ b/observability/prometheusrules.jsonnet
@@ -76,6 +76,8 @@ local appSREOverwrites(environment) = {
       else if
         std.startsWith(name, 'telemeter') then 'Tg-mH0rizaSJDKSADJ'
       else if
+        std.startsWith(name, 'loki_tenant') then 'f6fe30815b172c9da7e813c15ddfe607'
+      else if
         std.startsWith(name, 'loki') then 'Lg-mH0rizaSJDKSADX'
       else if
         std.startsWith(name, 'gubernator') then 'no-dashboard'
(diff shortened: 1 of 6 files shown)


observability/prometheusrules.jsonnet
```diff
@@ -139,13 +157,24 @@ local appSREOverwrites(environment) = {
...
+                    dashboardID(g.name).id,
+                    g.name,
+                    dashboardDatasource(environment).datasource,
+                    dashboardNamespace(environment, g.name).namespace,
+                  ],
+                }
               else
                 {
                   runbook: 'https://github.com/rhobs/configuration/blob/main/docs/sop/observatorium.md#%s' % std.asciiLower(r.alert),
-                  dashboard: 'https://grafana.app-sre.devshift.net/d/%s/%s?orgId=1&refresh=10s&var-datasource=%s&var-namespace={{$labels.namespace}}&var-job=All&var-pod=All&var-interval=5m' % [
```
kakkoyun, line 145: Shall we use grafana dashboard variables for namespaces?

kakkoyun: This is something we'd rather prefer to keep. Even more, we'd like to make dashboards/rules more dynamic using this. Because we use the same alerts and dashboards against multiple namespaces dynamically. We don't know what we're going to have in this repo. So alert templating kind of help for the links.

kakkoyun (changes requested):
I've some concerns commented inline.

observability/prometheusrules.jsonnet
periklis: ACK, i clearly missed the perspective here that these links are rendered in the alerts. I was looking on these links as being broken when I used them raw from the rendered YAML file.

kakkoyun (approved):
🚀

PR #15, opened 2021-05-25
services: fix the build

squat:
Right now the code in this repository does not compile when running
`make all -B`. This commit fixes the jsonnet so that it is able to
compile.


onprem (approved):
LGTM!

kakkoyun (commented):
💯

PR #18, opened 2021-05-27
Customize disabling down sampling for Thanos Compact component

spaparaju:
Customize disabling down sampling for Thanos Compact component. 
AFAIK as applying conditional logic at OpenShift templates is not possible, this PR directly applies the cmd line argument '--downsampling.disable' at Thanos Compact and bypass (not use) the corresponding JSonnet variable 'disableDownsampling' at Kube-Thanos.

### DIFF
--- a/services/observatorium-metrics-template-overwrites.libsonnet
+++ b/services/observatorium-metrics-template-overwrites.libsonnet
@@ -82,8 +82,7 @@ local oauthProxy = import './sidecars/oauth-proxy.libsonnet';
                   // Temporary workaround on high cardinality blocks for 2w.
                   // Since we have only 2w retention, there is no point in having 2w blocks.
                   // See: https://issues.redhat.com/browse/OBS-437
-                  args+: ['--debug.max-compaction-level=3'] +
-                         if disableDownsamplingFlag != null then disableDownsamplingFlag,
+                  args+: ['--debug.max-compaction-level=3'] + disableDownsamplingFlag,
                 } else c
                 for c in super.containers
               ],
--- a/services/observatorium-metrics-template-overwrites.libsonnet
+++ b/services/observatorium-metrics-template-overwrites.libsonnet
@@ -78,7 +78,7 @@ local oauthProxy = import './sidecars/oauth-proxy.libsonnet';
                   // Temporary workaround on high cardinality blocks for 2w.
                   // Since we have only 2w retention, there is no point in having 2w blocks.
                   // See: https://issues.redhat.com/browse/OBS-437
-                  args+: ['--debug.max-compaction-level=3'],
+                  args+: ['--debug.max-compaction-level=3', '--downsampling.disable=${THANOS_COMPACTOR_RETENTION_DISABLE_DOWNSAMPLING}'],
                 } else c
                 for c in super.containers
               ],
--- a/services/observatorium-metrics-template-overwrites.libsonnet
+++ b/services/observatorium-metrics-template-overwrites.libsonnet
@@ -69,6 +69,10 @@ local oauthProxy = import './sidecars/oauth-proxy.libsonnet';
       statefulSet+: {
         spec+: {
           replicas: '${{THANOS_COMPACTOR_REPLICAS}}',
+          local disableDownsamplingFlag = if compact.config.disableDownsampling then
+            ['--debug.max-compaction-level=3']
+          else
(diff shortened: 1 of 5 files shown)


services/observatorium-metrics-template-overwrites.libsonnet
```diff
@@ -78,7 +82,7 @@ local oauthProxy = import './sidecars/oauth-proxy.libsonnet';
                   // Temporary workaround on high cardinality blocks for 2w.
                   // Since we have only 2w retention, there is no point in having 2w blocks.
                   // See: https://issues.redhat.com/browse/OBS-437
-                  args+: ['--debug.max-compaction-level=3'],
+                  args+: kp,
```
squat, line 85: This doesn't quite take care of everything, I think. This concatenates two arrays together, so if someone still sets `disableDownsampling: true` then we'll have the flag twice again.

spaparaju: Tested with these 3 possibilities: 

- disableDownsampling (true) is present, overriding variable (--downsampling.disable) is NOT added to the args list
- - disableDownsampling (false) is present, kube-thanos does not add '--downsampling.disable', the overriding variable (--downsampling.disable) is added to the args list
- - disableDownsampling  is not present, the overriding variable (--downsampling.disable) is added to the args list

services/observatorium-metrics-template-overwrites.libsonnet
```diff
@@ -69,6 +69,13 @@ local oauthProxy = import './sidecars/oauth-proxy.libsonnet';
       statefulSet+: {
         spec+: {
           replicas: '${{THANOS_COMPACTOR_REPLICAS}}',
+          local debugArgs = ['--debug.max-compaction-level=3'],
```
squat, line 72: Maybe we don't need this extra variable and instead at the end we can do:

```jsonnet
args+: ['--debug.max-compaction-level=3'] + compactArgs,
```

squat (approved):
lgtm! one simplification but not a blocker

services/observatorium-metrics-template-overwrites.libsonnet
```diff
@@ -69,6 +69,10 @@ local oauthProxy = import './sidecars/oauth-proxy.libsonnet';
       statefulSet+: {
         spec+: {
           replicas: '${{THANOS_COMPACTOR_REPLICAS}}',
+          local disableDownsamplingFlag =
+            if !compact.config.disableDownsampling then
+              ['--downsampling.disable=${THANOS_COMPACTOR_RETENTION_DISABLE_DOWNSAMPLING}']
+            else [],
           template+: {
```
squat, line 76: I think this has a regression compared to the last version. What happens now if `compact.config.disableDownsampling` is true? No one appends the flag

spaparaju: If 'disableDownsampling' (Jsonnet variable of kube-thanos) is true, 'disableDownsamplingFlag' (the local JSonnet variable of RHOBS) is initialized with an empty array and 'disableDownsamplingFlag' (the local JSonnet variable of RHOBS) is not added ([another condition before args are added](https://github.com/spaparaju/configuration/blob/d5358bc893362bcd9fb9036afa831eeb0bb7ad8c/services/observatorium-metrics-template-overwrites.libsonnet#L86)) to the args. The original 'disableDownsampling' (Jsonnet variable of kube-thanos)  is already added to the args of the manifest.

squat: Ah `compact.config.disableDownsampling` is already handled by someone else? Oki 👍

services/observatorium-metrics-template-overwrites.libsonnet
```diff
@@ -78,7 +82,8 @@ local oauthProxy = import './sidecars/oauth-proxy.libsonnet';
                   // Temporary workaround on high cardinality blocks for 2w.
                   // Since we have only 2w retention, there is no point in having 2w blocks.
                   // See: https://issues.redhat.com/browse/OBS-437
-                  args+: ['--debug.max-compaction-level=3'],
+                  args+: ['--debug.max-compaction-level=3'] +
+                         if disableDownsamplingFlag != null then disableDownsamplingFlag,
```
squat, line 86: I done think we need this conditional, right? We can always consternate an empty array without any effect

spaparaju: Removed this condition and tested for all the possibilities (compact.config.disableDownsampling is not present / true / false)

PR #19, opened 2021-05-31
services: add configmap-reloader image to template

squat:
Right now, the CONFIGMAP_RELOADER_IMAGE parameters do not appear in the
OpenShift template for Observatorium metrics because they are declared
in the wrong file. This commit fixes this so that the template is valid.

Signed-off-by: Lucas Servén Marín <lserven@gmail.com>

cc @onprem @spaparaju


kakkoyun (approved):
LGTM :rocket:

PR #21, opened 2021-06-02
change the default value to disable downsampling

spaparaju:
The PR sets the default value for the libsonnet variable `THANOS_COMPACTOR_RETENTION_DISABLE_DOWNSAMPLING` as  `true`. This helps to disable downsampling as the default option.  
For specific cluster deployments, this variable `THANOS_COMPACTOR_RETENTION_DISABLE_DOWNSAMPLING` can be overridden (to 'false' to enable downsampling) at app-sre interface.


squat (approved):

squat:
Ah wait, looks like we need to commit the generated file

spaparaju:
Yes @squat , added changes to the generated template file.

PR #24, opened 2021-06-04
Monitor api memcached

squat:
Right now, the API memcached is not monitored because the service monitor does not have the correct namespace selector. This commit fixes this bug.

cc @kakkoyun @spaparaju


onprem (approved):
LGTM! Thanks for catching this.

PR #26, opened 2021-06-04
change as per the expected format of Thanos bool variables

spaparaju:
This PR updates the overwriting bool variable (for disable.downsampling) from the app-interface. This change is required as bool arguments Thanos cmd line does not accept the format of `--var=true/false`

### DIFF
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -2600,7 +2600,7 @@ parameters:
 - name: THANOS_COMPACTOR_RETENTION_RESOULTION_ONE_HOUR
   value: 1s
 - name: THANOS_COMPACTOR_RETENTION_DISABLE_DOWNSAMPLING
-  value: "--downsampling.disable"
+  value: --downsampling.disable
 - name: THANOS_COMPACTOR_MEMORY_LIMIT
   value: 5Gi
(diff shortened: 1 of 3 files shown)


resources/services/observatorium-metrics-template.yaml
```diff
@@ -123,7 +123,7 @@ objects:
           - --delete-delay=48h
           - --deduplication.replica-label=replica
           - --debug.max-compaction-level=3
-          - --downsampling.disable=${THANOS_COMPACTOR_RETENTION_DISABLE_DOWNSAMPLING}
+          - ${THANOS_COMPACTOR_RETENTION_DISABLE_DOWNSAMPLING}
```
squat, line 126: What do we set this to when we don't want to disable downsampling?

spaparaju: -  `--disable.downsampling` for disabling the downsampling (the default value at the config. repo)
- -  `--no-disable.downsampling` for enabling the downsampling (will be supplied from app-sre saas.yaml)

squat (approved):

resources/services/observatorium-metrics-template.yaml
squat: Ah, i didn't know about this feature

PR #32, opened 2021-06-08
Fix make validate

kakkoyun:
Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,12 @@ include .bingo/Variables.mk
 
 SED ?= sed
 XARGS ?= xargs
-OC ?= oc
+
+TMP_DIR := $(shell pwd)/tmp
+BIN_DIR ?= $(TMP_DIR)/bin
+OS ?= $(shell uname -s | tr '[A-Z]' '[a-z]')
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -113,3 +118,14 @@ clean:
 	find resources/observability/prometheusrules -type f ! -name '*.yaml' -delete
 	find resources/observability/grafana/observatorium -type f ! -name '*.yaml' -delete
 	find resources/observability/grafana/observatorium-logs -type f ! -name '*.yaml' -delete
+
+# Tools
+$(TMP_DIR):
```
squat, line 123: Looks like this and BIN_DIR don't appear as dependencies of anything, do this code won't get executed, no? Should we make it an ordering dependency of a target?

squat:
Amazing thank you :))

Makefile
kakkoyun: BIN_DIR used in OC target but you're right we're missing TMP_DIR adding it now.

kakkoyun: I guess it just worked because of `mkdir -p`

PR #33, opened 2021-06-10
Add Token Refresher to telemeter template

onprem:
This PR adds a new OpenShift template for deploying Token Refresher.

Using this template we can deploy a token-refresher instance pointing at the Observatorium API configured for a specific tenant. The token-refresher instance will be secured using the OpenShift OAuth Proxy.

We are planning to deploy one for the `telemeter` tenant so that we can migrate the users using the `infogw-*` URLs to go through Obs. API.


periklis (approved):
LGTM

Makefile
```diff
@@ -78,7 +78,7 @@ whitelisted_metrics: $(GOJSONTOYAML) $(GOJQ)
 
 .PHONY: manifests
 manifests: format $(VENDOR_DIR)
-manifests: resources/services/telemeter-template.yaml resources/services/jaeger-template.yaml resources/services/conprof-template.yaml
+manifests: resources/services/telemeter-template.yaml resources/services/jaeger-template.yaml resources/services/conprof-template.yaml resources/services/token-refresher-template.yaml
```
squat, line 81: Why does this need its own template? I would imagine it could be part of Telemeter dice it's for the infogw URLs

resources/services/token-refresher-template.yaml
```diff
@@ -0,0 +1,169 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: token-refresher
+objects:
+- apiVersion: apps/v1
+  kind: Deployment
```
squat, line 7: In general we try to ensure every workload had a service monitor before being deployed. Can we add one to the obs/obs repo?

Makefile
onprem: Yeah that makes sense. I added a new template because 1) it makes it easier to remove this in the future, and 2) this can be used with any other tenant as well (with observatorium-mst) if needed.

onprem: Moved it to the telemeter template.

kakkoyun (approved):
LGTM 🚀

bwplotka (commented):
Some suggestion from our team sync:

* Can we use the telemeter namespace and tie this to telemeter?
* I am afraid we would need to do the same for MST at some point 🤔  (let'sthink later)

onprem:
> Can we use the telemeter namespace and tie this to telemeter?

Yep, that's what we are doing now. I have made this a part of the telemeter-template.

PR #40, opened 2021-07-01
Add tenant ID to rules

squat:
Now that the Observatorium API enforces tenant isolation at query time,
we need to ensure that all recording rules include the tenant ID as a
label so that the time series produced by the rules remain query-able.

### DIFF
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -1249,38 +1249,58 @@ objects:
   data:
     observatorium.yaml: |-
       "groups":
-      - "interval": "3m"
-        "name": "observatorium"
+      - "interval": "1m"
+        "name": "telemeter-telemeter.rules"
         "rules":
         - "expr": |
             count by (name,reason) (cluster_operator_conditions{condition="Degraded"} == 1)
+          "labels":
+            "tenant_id": "FB870BF3-9F3A-44FF-9BF7-D7A047A52F43"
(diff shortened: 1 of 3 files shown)


resources/services/observatorium-metrics-template.yaml
```diff
@@ -1249,38 +1249,58 @@ objects:
...
+          "labels":
+            "tenant_id": "FB870BF3-9F3A-44FF-9BF7-D7A047A52F43"
           "record": "id_version_ebs_account_internal:cluster_subscribed"
         - "expr": |
             0 * (max by (_id,host_type) (topk by (_id) (1, label_replace(label_replace(label_replace(label_replace(label_replace(label_replace(cluster:virt_platform_nodes:sum, "host_type", "$1", "type", "(aws|ibm_.*|ovirt|none|rhev|gcp|openstack|hyperv|vmware)"), "host_type", "virt-unknown", "host_type", ""), "host_type", "kvm-unknown", "type", "kvm"), "host_type", "xen-unknown", "type", "xen.*"), "host_type", "metal", "host_type", "none"), "host_type", "ibm-$1", "host_type", "ibm[_-](power|systemz).*"))) or on(_id) label_replace(max by (_id) (cluster_version{type="current"}), "host_type", "", "host_type", ""))
```
bwplotka, line 1281: I assume we don't need to put tenant_Id in select, because it will be appended by default if not present?

bwplotka (approved):
LGTM! Thanks

PR #44, opened 2021-07-19
Add tracing endpoints to opa-ams config & regen templates

ianbillett:
This PR follows the example set [here](https://github.com/observatorium/api/pull/127/files) and adds tracing context to the opa-ams sidecar. 

### DIFF
--- a/services/sidecars/opa-ams.libsonnet
+++ b/services/sidecars/opa-ams.libsonnet
@@ -77,6 +77,34 @@ function(params) {
                 for tenant in std.objectFields(opa.config.mappings)
               ]
             else []
+          ) + (
+            if std.objectHas(opa.config.internal, 'tracing') then
+              [] + (
+                if std.objectHas(opa.config.internal.tracing, 'endpoint') then
+                  [
+                    '--internal.tracing.endpoint=' + opa.config.internal.tracing.endpoint,
+                  ]
+                else []
+              ) + (
+                if std.objectHas(opa.config.internal.tracing, 'endpointType') then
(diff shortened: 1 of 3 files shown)


services/sidecars/opa-ams.libsonnet
```diff
@@ -77,6 +77,34 @@ function(params) {
...
+                  [
+                    '--internal.tracing.sampling-fraction=' + opa.config.internal.tracing.samplingFraction,
+                  ]
+                else []
+              ) + (
+                if std.objectHas(opa.config.internal.tracing, 'serviceName') then
+                  [
+                    '--internal.tracing.service-name=' + opa.config.internal.tracing.serviceName,
+                  ]
+                else []
```
periklis, line 105: Do we have defaults for these config options?

onprem (approved):
LGTM!

services/sidecars/opa-ams.libsonnet
onprem: Yes, these three opa-ams flags have default values. ref: https://github.com/observatorium/opa-ams/blob/master/main.go#L113-L120

ianbillett: The default options for each of these flags is implemented in the `opa-ams` binary itself [here](https://github.com/observatorium/opa-ams/blob/master/main.go#L113-L120). The semantics here are "if the value has been specified in opa.config then add it to the jsonnet with the right flag, else leave it unspecified and use the default from the binary".

PR #46, opened 2021-07-21
Make target namespace configurable for ams remote write proxy

onprem:
The AMS remote write proxy writes directly to Thanos Receive, which will move to a different namespace after migration. This PR changes the target namespace to `${OBSERVATORIUM_METRICS_NAMESPACE}` so that it continues to work after migration.


ianbillett (approved):
LGTM

PR #48, opened 2021-07-26
configuration/telemeter/metrics.json: Refresh allow list

lilic:
Hello all 👋 I ran `make whitelisted_metrics` and then `make`. The generation seems to have touched other files, not sure if that is intended, but I only committed the allow list.

Side note: Also `make clean` doesn't seem to actually clear the state for me.


onprem:
Heya :wave: 

> I ran `make whitelisted_metrics` and then `make`. The generation seems to have touched other files, not sure if that is intended, but I only committed the allow list.

I tried this on the current main but couldn't reproduce. In my case it only updated the metrics.json first and then the template got updated as expected, (I had to use `make --always-make all` though, because somehow it was not picking up that metrics.json got updated).

> Side note: Also `make clean` doesn't seem to actually clear the state for me.

The target is used to only deleting the json files generated from jsonnet, after converting them to yaml using gojsontoyaml. I guess we can name it better because `make clean` is generally used of cleaning up state as you expected.

onprem (approved):
Thanks, LGTM!

lilic:
Thanks Prem for the explanation!

PR #49, opened 2021-07-29
configuration/telemeter: track number of active alertmanager integrations

paulfantom:
Allow-list data from https://github.com/openshift/cluster-monitoring-operator/pull/1209


paulfantom:
Side note: docs about adding telemetry metrics say to run `make whitelisted_metrics && make` which is **not** regenerating files in `resources/` when `bingo` is already in latest version.

ianbillett (approved):
LGTM

PR #50, opened 2021-08-01
Update observatorium-logs-template.yaml

eranra:
Use parameter SERVICE_ACCOUNT_NAME for all service accounts


periklis (changes requested):
This is change should be reflected in the jsonnet code and not in the final generated yaml.

eranra:
@periklis which files == ?

PR #52, opened 2021-08-27
telemeter: report Prometheus scrape related metrics

arajkumar:
Regenerate config to pickup metrics exposed as part of https://github.com/openshift/cluster-monitoring-operator/pull/1313

Signed-off-by: Arunprasad Rajkumar <arajkuma@redhat.com>


onprem:
The `telemeter-template.yaml` needs to be regenerated because you updated the `metrics.json` file. You have to run `make manifests` to do that,  this will also make the CI pass.

bwplotka (approved):
Thanks!

PR #55, opened 2021-09-22
Fix Thanos latency runbook, add requirements

jessicalins:
Just fixed some broken links and updated the alert severity (as defined [here](https://github.com/rhobs/configuration/blob/main/resources/observability/prometheusrules/observatorium-thanos-production.prometheusrules.yaml#L495))


bwplotka (approved):
Thank you! LGTM

PR #59, opened 2021-09-28
Add missing NAMESPACES parameter

onprem:
The `NAMESPACES` parameter was missing from the metric-federation ruler template, causing problems when deploying.


matej-g (approved):
LGTM!

PR #61, opened 2021-09-28
chore: Update federation rule kafka metric names to include gibibytes

davidffrench:
Updated kafka federation metric names to include gibibytes


davidffrench:
Once the corresponding name changes have been rolled out in our environment, I will remove the WIP from this PR.

ianbillett (approved):
Thanks @davidffrench :raised_hands: This is not automatically rolled out so we will have to wait for someone to do this tomorrow.

PR #62, opened 2021-09-29
Update alerts and dashboards for metric federation Ruler

onprem:
Updated the Thanos Rule alerts and dashboard for metric-federation Ruler.


bwplotka (approved):
LGTM! (:

PR #63, opened 2021-09-30
Conprof/MST template: Fix namespace, service account

jessicalins:
Refs https://issues.redhat.com/browse/MON-1934

Not sure if we should update the service account as well?
I also didn't get how we update this in staging (I guess once we update the commit ref in app-interface directly?)

Signed-off-by: Jéssica Lins <jlins@redhat.com>

### DIFF
--- a/services/conprof-template.jsonnet
+++ b/services/conprof-template.jsonnet
@@ -341,9 +341,9 @@ local conprof = c + c.withConfigMap {
     ],
     parameters: [
       { name: 'IMAGE_TAG', value: 'master-2020-04-29-73bf4f0' },
-      { name: 'NAMESPACE', value: 'telemeter' },
+      { name: 'NAMESPACE', value: 'observatorium' },
       { name: 'OBSERVATORIUM_MST_NAMESPACE', value: 'observatorium-mst' },
-      { name: 'SERVICE_ACCOUNT_NAME', value: 'prometheus-telemeter' },
+      { name: 'SERVICE_ACCOUNT_NAME', value: 'observatorium' },
     ],
   },
 }
--- a/services/conprof-template.jsonnet
+++ b/services/conprof-template.jsonnet
@@ -324,7 +324,7 @@ local conprof = c + c.withConfigMap {
     ],
     parameters: [
       { name: 'IMAGE_TAG', value: 'master-2020-04-29-73bf4f0' },
-      { name: 'NAMESPACE', value: 'telemeter' },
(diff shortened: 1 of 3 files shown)


services/conprof-template.jsonnet
```diff
@@ -324,7 +324,7 @@ local conprof = c + c.withConfigMap {
     ],
     parameters: [
       { name: 'IMAGE_TAG', value: 'master-2020-04-29-73bf4f0' },
-      { name: 'NAMESPACE', value: 'telemeter' },
+      { name: 'NAMESPACE', value: 'observatorium' },
       { name: 'OBSERVATORIUM_LOGS_NAMESPACE', value: 'observatorium-logs' },
       { name: 'SERVICE_ACCOUNT_NAME', value: 'prometheus-telemeter' },
```
jessicalins, line 329: not sure if service account needs to be changed here. Also in https://github.com/rhobs/configuration/blob/399d551ca024450dcb78075e6ab86613517e3d06/services/conprof-template.jsonnet#L346 and in `conprof-observatorium-logs-rbac-template.yaml` and `conprof-observatorium-mst-rbac-template.yaml` (?)

matej-g:
We can update the values here as well, but I think these will be actually overridden by the values from `app-interface` (https://gitlab.cee.redhat.com/service/app-interface/blob/master/data/services/rhobs/observatorium-mst/cicd/saas.yaml#L300 and https://gitlab.cee.redhat.com/service/app-interface/blob/master/data/services/rhobs/observatorium/observatorium-logs/cicd/saas.yaml#L201 respectively for both MST and logs). I think we should update this in `app-interface` in order for it to take effect.

onprem (approved):
Just as Matej said, this is not required because we are going to override this anyways. But I think the values serve kinda as a hint/documentation for people looking at the templates, so let's definitely merge this. Thanks!

services/conprof-template.jsonnet
onprem: Well yes, this value will be `observatoium` in upstream. No need to update here because we are going to override this in app-interface anyways.

jessicalins: ah interesting, thanks for the explanation!
so this means that not all parameters in this repo are in sync with the ones in app-interface? I find it a bit confusing. I think would be better to keep both in sync, but at the same time this means updating in 2 places..
I guess the actual values maybe don't matter that much, but in case we add new parameters/update would be nice to keep them in sync with app-interface too

PR #64, opened 2021-10-04
Conprof: Update service account

jessicalins:
Just as documentation, so we keep it consistent to what is actually being deployed in `app-interface`
Context: https://issues.redhat.com/browse/MON-1934


onprem (approved):
LGTM, thanks!

PR #65, opened 2021-10-04
MON-1390: Add avalanche deployment that writes data to the RHOBS tenant

ianbillett:
This PR adds a deployment of [open-fresh/avalanche](https://github.com/open-fresh/avalanche) that will run in both MST & Telemeter instances of RHOBS.

The purpose of this is to write a **known** & **static** amount of data to our tenant, so that we can make semi-precise measurements of query path performance for our SLI [implementation](https://issues.redhat.com/browse/MON-1390).

As the comments say, this deployment will write 1M samples per hour - 8.3k series at 30s intervals.

Tested this configuration _locally_, but not in a cluster yet. In fact - I'm not sure how we will even query this data quite yet.

### DIFF
--- a/resources/services/observatorium-template.yaml
+++ b/resources/services/observatorium-template.yaml
@@ -541,7 +541,7 @@ objects:
           - --value-interval=60
           - --series-interval=86400
           - --metric-interval=86400
-          - --const-label=tenant_id="770c1124-6ae8-4324-a9d4-9ce08590094b"
+          - --const-label=tenant_id="0fc2b00e-201b-4c17-b9f2-19d91adc4fd2"
           image: quay.io/freshtracks.io/avalanche:master-2020-12-28-0c1c64c
           name: avalanache-remote-writer
 - apiVersion: v1
--- a/resources/services/observatorium-template.yaml
+++ b/resources/services/observatorium-template.yaml
@@ -509,6 +509,41 @@ objects:
         app.kubernetes.io/instance: observatorium
         app.kubernetes.io/name: gubernator
         app.kubernetes.io/part-of: observatorium
+- apiVersion: apps/v1
+  kind: Deployment
+  metadata:
+    labels:
+      app.kubernetes.io/component: avalanche
+      app.kubernetes.io/name: avalanche-remote-writer
+      app.kubernetes.io/part-of: observatorium
+    name: avalanache-remote-writer
+  spec:
+    replicas: 1
+    selector:
+      matchLabels:
+        app.kubernetes.io/component: avalanche
+        app.kubernetes.io/name: avalanche-remote-writer
+        app.kubernetes.io/part-of: observatorium
+    template:
+      metadata:
+        labels:
+          app.kubernetes.io/component: avalanche
+          app.kubernetes.io/name: avalanche-remote-writer
+          app.kubernetes.io/part-of: observatorium
+      spec:
+        containers:
+          args:
+          - --metric-count=8333
+          - --series-count=1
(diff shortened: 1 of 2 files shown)


services/observatorium.libsonnet
```diff
@@ -361,6 +361,59 @@ local memcached = (import 'github.com/observatorium/observatorium/configuration/
...
+        template: {
+          metadata: {
+            labels: config.commonLabels,
+          },
+          spec: {
+            containers: {
+              name: config.name,
+              image: config.image,
+              args: [
+                '--metric-count=1',  // we only get one metric __name__
```
jessicalins, line 396: I think would be nice to also add an entry in the README about this

services/observatorium.libsonnet
```diff
@@ -361,6 +361,59 @@ local memcached = (import 'github.com/observatorium/observatorium/configuration/
...
+                '--series-count=8333',  // this is set so that we write 1M samples per hour to our test tenant
+                '--remote-url="http://%s.%s.svc.cluster.local:%d"' % [
+                  obs.thanos.receiversService.metadata.name,
+                  obs.config.namespaces.metrics,
+                  obs.thanos.receiversService.spec.ports[2].port,
```
jessicalins, line 401: would we have the case where [2] is not accessible? (or with a different value)

resources/services/observatorium-template.yaml
```diff
@@ -509,6 +509,41 @@ objects:
...
+          args:
+          - --metric-count=1
+          - --series-count=8333
+          - --remote-url="http://observatorium-thanos-receive.${OBSERVATORIUM_METRICS_NAMESPACE}.svc.cluster.local:19291"
+          - --remote-write-interval=30s
+          - --remote-requests-count=1000000
+          - --value-interval=60
+          - --series-interval=86400
+          - --metric-interval=86400
+          - --const-label=tenant_id="770c1124-6ae8-4324-a9d4-9ce08590094b"
```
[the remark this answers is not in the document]
ianbillett: `--label-count` defaults to 10 which is a reasonable default - so no requirement to change :+1:

services/observatorium.libsonnet
ianbillett: Which README do you mean?

services/observatorium.libsonnet
ianbillett: This snippet is taken verbatim from [here](https://github.com/rhobs/configuration/blob/d63616f25fb28f08770a26c432175534de12d35f/services/observatorium.libsonnet#L179-L182) - if [2] is not accessible, then the API server will be in big trouble :)

services/observatorium.libsonnet
jessicalins: just the main one: https://github.com/rhobs/configuration

ianbillett: Without a precedent from other services (i.e. Up, API, Thanos) - I don't want to unnecessarily clutter the main readme if that makes sense?

I don't think our documentation is _great_ by any means :sweat_smile: there's just no pattern to follow at the moment.

jessicalins: I just thought since this will frequently write static data to our tenant, would be nice to refer back to it (I think just adding an entry with the same content of the description of this PR would already be enough) - specially if we don't have great documentation, is one more reason to start improving it :) but not a blocker from my side, just a suggestion

matej-g (approved):
Let's (ava)launch this :mountain_snow:

PR #67, opened 2021-10-04
Add queries to measure read path SLIs

ianbillett:
This PR updates the existing Up binary deployment to make queries that measure the read path performance for our [SLOs](https://docs.google.com/document/d/1wJjcpgg-r8rlnOtRiqWGv0zwr1MB6WwkQED1XDWXVQs/edit).

The existing queries have been [untouched](https://github.com/rhobs/configuration/commit/bfc2bdf2c045c8708b98a3ccdb29decf64429172) for > 1 year - even then they were ported from elsewhere. I don't know _why_ we have them, please shout if you do :)

Until https://github.com/rhobs/configuration/pull/65 PR is merged and shipped, there will be no data to query, however it is harmless to query for missing data.

The hard-coded default of `up` is to execute the custom queries serially with 100ms delay between each.

### DIFF
--- a/configuration/observatorium/queries.libsonnet
+++ b/configuration/observatorium/queries.libsonnet
@@ -1,104 +1,16 @@
 {
   queries: [
     {
-      name: 'Clusters',
-      query: 'avg_over_time(sum(count by (_id) (max without (prometheus,receive,instance) ( cluster_version{type="current"} )) + on (_id) group_left(_blah) (topk by (_id) (1, 0 *subscription_labels{email_domain!~"redhat.com|(^|.*\\\\.)ibm.com"})))[7d:12h])',
+      name: 'query-path-sli-1M-samples',
+      query: 'avg_over_time(avalanche_metric_mmmmm_0_0{tenant_id="1610b0c3-c509-4592-a256-a1871353dbfa"}[1h])',
     },
     {
-      name: 'Clusters aged 1w',
-      query: 'avg_over_time(sum(count by (_id) (max without (prometheus,receive,instance) ( (time() - cluster_version{type="initial"}) > (7 * 24 * 60 * 60) )) + on (_id) group_left(_blah) (topk by (_id) (1, 0 *subscription_labels{email_domain!~"redhat.com|(^|.*\\\\.)ibm.com"})))[7d:12h])',
+      name: 'query-path-sli-10M-samples',
+      query: 'avg_over_time(avalanche_metric_mmmmm_0_0{tenant_id="1610b0c3-c509-4592-a256-a1871353dbfa"}[10h])',
     },
     {
-      name: 'Nodes',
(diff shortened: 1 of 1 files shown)


configuration/observatorium/queries.libsonnet
```diff
@@ -1,104 +1,16 @@
 {
   queries: [
     {
-      name: 'Clusters',
-      query: 'avg_over_time(sum(count by (_id) (max without (prometheus,receive,instance) ( cluster_version{type="current"} )) + on (_id) group_left(_blah) (topk by (_id) (1, 0 *subscription_labels{email_domain!~"redhat.com|(^|.*\\\\.)ibm.com"})))[7d:12h])',
+      name: 'query-path-sli-1M-samples',
+      query: 'avg_over_time(avalanche_metric_mmmmm_0_0{tenant_id="1610b0c3-c509-4592-a256-a1871353dbfa"}[1h])',
```
ianbillett, line 5: Given the config [here](https://github.com/rhobs/configuration/pull/65/files#diff-d2332f4f83480e6f83ef074dfb60f41294a4cb2ee1662f237f6545bfc5cdfe3dR536-R537), we know the metric name will always be `avalanche_metric_mmmmm_0_0`

spaparaju: would the mentioned `tenant_id` (`"1610b0c3-c509-4592-a256-a1871353dbfa`) be the same for metric `avalanche_metric_mmmmm_0_0` at both telemeter and MST tenants / need to parameterize the `tenant_id` for different RHOBS tenants ?

ianbillett: Ah good call out - I hadn't checked.

ianbillett: Ok so slight problem here:
* Telemeter instance, rhobs tenant_id = `770c1124-6ae8-4324-a9d4-9ce08590094b` [link](https://vault.devshift.net/ui/vault/secrets/app-interface/show/app-sre-stage/telemeter-stage/observatorium-observatorium-api)
* MST instance, osd tenant_id = `770c1124-6ae8-4324-a9d4-9ce08590094b`  [link](https://vault.devshift.net/ui/vault/secrets/app-interface/show/app-sre-stage/telemeter-stage/observatorium-observatorium-mst-api)

I think when we created the mst instance we copypasta'd the original tenants yaml file.

I think we need to generate a new UUID for the rhobs tenant and add this to both the telemeter & mst instance tenant yaml files.

ianbillett: Ok I've generated a new uuid for the rhobs tenant, and will update the `tenants.yaml` files in telemeter & mst instances.

This is non-blocking for this PR though.

matej-g (approved):
Looks good. I'm also not sure where the existing queries are coming from, but it seems like the proposed queries will gives us a much better idea on how we're performing on the query side :+1:.

PR #70, opened 2021-10-06
Fix crashloop by unquoting remote-url 

ianbillett:
The avalanche deployment is [crashlooping](https://console-openshift-console.apps.app-sre-stage-0.k3s7.p1.openshiftapps.com/k8s/ns/observatorium-stage/deployments/avalanache-remote-writer) due to a url parsing error:
```
error: invalid URL: parse "\"http://observatorium-thanos-receive.observatorium-metrics-stage.svc.cluster.local:19291/api/v1/receive\"": first path segment in URL cannot contain colon
```
See [this](https://play.golang.org/p/MIYJEdAEM6G) Go playground for an example.

This PR fixes this problem, by removing the inner quotes on the URL. 

### DIFF
--- a/resources/services/observatorium-template.yaml
+++ b/resources/services/observatorium-template.yaml
@@ -535,7 +535,7 @@ objects:
         - args:
           - --metric-count=1
           - --series-count=8333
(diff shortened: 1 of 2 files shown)


jessicalins (approved):
:watermelon:

PR #76, opened 2021-10-07
Make Avalanche append the correct tenant headers

ianbillett:
Currently, our deployment of avalanche is not working as expected. This is because `Thanos Receive` does not respect `tenant_id`s supplied in metric label sets. Thought process [here](https://coreos.slack.com/archives/C026ZQSB468/p1633618569174300).

To fix this, we need to supply `THANOS-TENANT` in the `remote_write` request headers. I've made the changes and raised https://github.com/open-fresh/avalanche/pull/25 to make this configurable in the upstream.

However, since that repo was last updated last December, I've built the dockerfile locally and pushed it to the Observatorium quay [repo](https://quay.io/repository/observatorium/avalanche?tab=tags). We will switch the images if / when the upstream PR is merged.

Tested locally and things are working as expected.

Finally, I've updated the value interval from 60s to 1h. Random gauges changing every 60s is not representative of real-world users.

### DIFF
--- a/resources/services/observatorium-template.yaml
+++ b/resources/services/observatorium-template.yaml
@@ -553,11 +553,12 @@ objects:
           - --remote-url=http://observatorium-thanos-receive.${OBSERVATORIUM_METRICS_NAMESPACE}.svc.cluster.local:19291/api/v1/receive
           - --remote-write-interval=30s
           - --remote-requests-count=1000000
-          - --value-interval=60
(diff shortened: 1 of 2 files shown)


matej-g (approved):
Changes in the upstream are looking good, acceptable as a workaround :+1: Let's see if the upstream repo is still actively maintained, if not I would consider forking it.

PR #89, opened 2021-10-25
Add pullSecret for quay.io in SA so it's able to get image from priva…

OliviaHY:
### Why:
We will be making quay.io/app sre/gubernator private and use a recently mirrored memcached repo which is private as well. 

### What:
Add pullSecret for quay.io in SA so it's able to get image from private repo.

### DIFF
--- a/services/telemeter.libsonnet
+++ b/services/telemeter.libsonnet
@@ -130,7 +130,9 @@
         },
       },
     },
-
+    serviceAccount+: {
+      imagePullSecrets+: [{ name: 'quay.io' }],
+    },
     serviceMonitor+: {
       spec+: {
         jobLabel: 'app.kubernetes.io/component',
--- a/jsonnetfile.lock.json
+++ b/jsonnetfile.lock.json
@@ -152,8 +152,8 @@
           "subdir": "jsonnet/telemeter"
         }
       },
-      "version": "0bea9a2c9711b8877f9c7b4b7aa5c0a55793d19d",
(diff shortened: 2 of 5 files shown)


services/telemeter.libsonnet
```diff
@@ -130,7 +130,9 @@
         },
       },
     },
-
+    serviceAccount+: {
+      imagePullSecrets+: [{ name: 'quay.io' }],
+    },
```
onprem, line 135: Looks like this is not working as intended. The telemeter memcached [does not have any service account configured](https://github.com/openshift/telemeter/blob/master/jsonnet/telemeter/server/kubernetes.libsonnet#L197) in it. This is why it is generating an invalid K8s object and the CI is failing.

I think the solution is that we give the Memcached it's own service account, and then we configure pull secrets for it.

onprem: I have created a PR for adding a service account for Memcached openshift/telemeter#403. Once it gets merged, we can update the dependency here and regenerate the manifests. CI should pass after this.

OliviaHY: That makes sense, thank you for helping @onprem! Is this not an issue with observatorium? Is this [repo](https://github.com/observatorium/observatorium/tree/main/website) where that can be checked?

OliviaHY: @onprem I saw the PR for adding the SA is merged, how do we update the dependency here?

onprem (approved):
This is looking good to go. Thanks @OliviaHY!

OliviaHY:
@periklis I think the changes you quested has been made. Please take another look.

periklis (approved):
LGTM

PR #93, opened 2021-11-04
Add SLO Dashboards

ianbillett:
This PR:
* Implements a Jsonnet library that generates SLO dashboards.
* Generates 4 SLO dashboards using this library that communicate our [SLOs](https://docs.google.com/document/d/1wJjcpgg-r8rlnOtRiqWGv0zwr1MB6WwkQED1XDWXVQs/edit) 

This is explicitly a _starting point_ that we will iterate from. If we believe that these SLOs are not communicating a representative user experience then we will revisit and update these SLOs.

Dashboards are available to preview in staging:
* [Telemeter Production SLOs](https://grafana.stage.devshift.net/d/f9fa7677fb4a2669f123f9a0f2234b47/telemeter-production-slos?orgId=1)
* [Telemeter Staging SLOs](https://grafana.stage.devshift.net/d/080e53f245a15445bdf777ae0e66945d/telemeter-staging-slos?orgId=1)
* [MST Production SLOs](https://grafana.stage.devshift.net/d/283e7002d85c08126681241df2fdb22b/mst-production-slos?orgId=1)
* [MST Staging SLOs](https://grafana.stage.devshift.net/d/92520ea4d6976f30d1618164e186ef9b/mst-stage-slos?orgId=1)

NB: If staging Grafana gets restarted the above will be unavailable to be viewed.


bwplotka (changes requested):
This is great, but can we actually leverage libsonnet and have one libsonnet dashboard for this SLO graphs and use variables to switch between environments and instances? It would be way more maintainable.  🤗 

Otherwise it looks super great!

jessicalins (commented):
cool! :taco: 
I would suggest also to update the descriptions of the SLOs that calculate the percentiles to `90th percentile of` instead of `90% of`

bwplotka (approved):
Great job. I think increment should look similar to this https://github.com/google/jsonnet/blob/master/test_suite/invariant.jsonnet#L40 but we can do that later I  guess. 

Thanks for this! It looks good and readable to me (:

PR #98, opened 2021-11-16
Alerts: Add alert for skipped misconfigured tenants

matej-g:

We recently introduced a new metric to ensure we track the number of tenants, which failed to be registered in the API due to wrong configuration (https://github.com/observatorium/api/pull/187).

This alert is a follow up on that change, to ensure we are aware of any misconfiguration immediately.

It also removes the exception for RHOBS tenant, which is now correctly configured in all environments.


bwplotka (approved):
LGTM!

Thanks!

observability/prometheusrules.jsonnet
```diff
@@ -540,10 +540,25 @@ local renderAlerts(name, environment, mixin) = {
               annotations: {
                 message: 'Increase in failed attempts to register with OIDC provider for {{ $labels.tenant }}',
               },
-              // TODO(@matej-g): The exclusion of 'rhobs' tenant should be removed
-              // after https://issues.redhat.com/browse/MON-1849 is resolved.
               expr: |||
-                sum(increase(observatorium_api_tenants_failed_registrations{tenant!="rhobs"}[5m])) by (tenant) > 0
+                sum(increase(observatorium_api_tenants_failed_registrations[5m])) by (tenant) > 0
```
moadz, line 544: What's the rationale for including the `rhobs` tenant?

docs/sop/observatorium.md
```diff
@@ -171,6 +171,32 @@ There is either outage on the side of the OIDC provider or the tenant authorizat
 
 ---
 
+## ObservatoriumTenantsSkippedDuringConfiguration
```
moadz, line 174: In this particular case it's explicitly OIDC configuration right? Is it useful to include that in the alert itself?

moadz (commented):
LGTM

observability/prometheusrules.jsonnet
matej-g: When we introduced the alert, the `rhobs` tenant (which is our team's own tenant intended for manual testing) was not configured properly, meaning we would be getting alerted on it all the time. Now that the tenant works properly, we can remove the exception.

docs/sop/observatorium.md
matej-g: In this case it can be actually any tenant configuration, whether authentication or authorization, we basically fail in the configuration-checking phase, so this occurs before a failure to reach OIDC would occur.

observability/prometheusrules.jsonnet
moadz: Cool thank you!

PR #102, opened 2021-11-22
Update allow list with nto_custom_profiles:count

dagrayvid:
Allow list new metric added in https://github.com/openshift/cluster-monitoring-operator/pull/1484.

Note:
Running `make all` also touched several other files (removing resources/observability/*prometheusrule.yaml). I left these changes out, please let me know if these should be added back in.


matej-g (approved):
Looks good, thanks!

The removed rules might have been just an intermittent thing - `make all` first removes and then regenerates the rules, so it appears as if for a moment they were deleted.

dagrayvid:
Okay. Thanks @matej-g.

PR #103, opened 2021-11-24
Observatorium Metrics: Provide extra store(s) via file SD

matej-g:
We currently use the `observatorium-template` instance for both our "core" Observatorium instance and the MST Observatorium instance. However, there is an important difference between the instances, introduced recently due to the metric federation ruler change (https://github.com/rhobs/configuration/pull/56) - this ruler exists only in the "core" instance and not in the MST instance.

Trying to pass this ruler address to querier in the MST instance is causing a repeated error in logs:
```
level=error ts=2021-11-24T14:37:54.232621974Z caller=resolver.go:99 msg="failed to lookup SRV records" host=_grpc._tcp.observatorium-thanos-metric-federation-rule.observatorium-mst-stage.svc.cluster.local err="no such host"
```
This is less than ideal since the querier tries to resolve the non-existing address roughly every 30 seconds.

Since we're unable to differentiate stores between the instance with templating, the workaround here is to provide the extra store(s) via file SD. The change means:
- Creating and mounting new config map volume with SD file in YAML format
- Passing the `store.sd-files` flag to querier
- The content of the file can be dynamically changed with a parameter (`THANOS_QUERIER_FILE_SD_TARGETS`), 
- By default, the targets array in the SD file is empty, meaning no additional stores are added
- Keep stores that are shared between instances in the `store` flags

### DIFF
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -234,6 +234,17 @@ objects:
           requests:
             storage: ${THANOS_COMPACTOR_PVC_REQUEST}
         storageClassName: ${STORAGE_CLASS}
+- apiVersion: v1
+  data:
+    file_sd.yaml: '- targets: ${THANOS_QUERIER_FILE_SD_TARGETS}'
+  kind: ConfigMap
+  metadata:
+    annotations:
+      qontract.recycle: "true"
+    labels:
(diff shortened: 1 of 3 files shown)


jessicalins (approved):
as far as I understand, looks good :)

onprem (approved):
LGTM as well!

PR #104, opened 2021-11-26
Add Thanos Query Frontend dashboard

jessicalins:
- Updates `thanos-mixin` dependency to get the latest `main` from the Thanos repo/mixin folder (I ran into some issues to update the dependency by using `jb update <repo_url_here>` - so after some research I found that this is an open issue in `jsonnet-bundler` (Updated the README that has a link to a workaround for it)
- Imports `query_frontend.libsonnet` to generate Query Frontend Grafana dashboard
- Runs `make grafana` (This also updated other Thanos dashboards, I think because now we updated the `thanos-mixin` version

### DIFF
--- a/resources/observability/prometheusrules/observatorium-thanos-stage.prometheusrules.yaml
+++ b/resources/observability/prometheusrules/observatorium-thanos-stage.prometheusrules.yaml
@@ -316,7 +316,7 @@ spec:
       for: 5m
       labels:
         service: telemeter
-        severity: medium
+        severity: high
     - alert: ThanosReceiveHighHashringFileRefreshFailures
       annotations:
         dashboard: https://grafana.app-sre.devshift.net/d/916a852b00ccc5ed81056644718fa4fb/thanos-receive?orgId=1&refresh=10s&var-datasource=app-sre-stage-01-prometheus&var-namespace={{$labels.namespace}}&var-job=All&var-pod=All&var-interval=5m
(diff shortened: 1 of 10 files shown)


resources/observability/prometheusrules/observatorium-thanos-stage.prometheusrules.yaml
```diff
@@ -362,6 +362,23 @@ spec:
       labels:
         service: telemeter
         severity: high
+    - alert: ThanosReceiveTrafficBelowThreshold
```
jessicalins, line 365: to fix this https://github.com/rhobs/configuration/runs/4336409317?check_suite_focus=true I had to run `make prometheusrules` commit: https://github.com/rhobs/configuration/pull/104/commits/8d681f6b54c61b6340134296cb286c9f175f7967

matej-g: Yeah, looks like we haven't updated this in some time, it looks like this was added to Thanos mixins some time ago already

matej-g (approved):
Nice one figuring out the bundler issue :clap: Looks good! Let's make sure the rules are updated as well once we'll be deploying the changes.

matej-g:
What is the status of the adding alerts to AppSRE Interface? I see one MR has been closed (https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/29529) but I don't think there has been any follow up after that.

jessicalins:
@matej-g good point, I think first we need to add a proper runbook for this alert here https://github.com/rhobs/configuration/blob/main/docs/sop/observatorium.md#thanosreceivetrafficbelowthreshold
I created this issue here for now: https://github.com/rhobs/configuration/issues/115
After we add the runbook here I think then we can add it to AppSRE.. What do you think?

PR #105, opened 2021-11-29
Prometheus Rules: Add tenant misconfiguration rule test

matej-g:

Adding test for recently introduced alerting rule - `ObservatoriumTenantsSkippedDuringConfiguration` (was asked for by AppSRE before it's rolled out).

### DIFF
--- a/observability/prometheus_rule_tests/observatorium-tenants.prometheusrulestests.yaml
+++ b/observability/prometheus_rule_tests/observatorium-tenants.prometheusrulestests.yaml
@@ -1,0 +1,53 @@
+---
+$schema: /app-interface/prometheus-rule-test-1.yml
+
+rule_files:
+# For upstream:
+# - /observability/prometheusrules/observatorium-tenants-stage.prometheusrules.yaml
+- /observability/prometheusrules/observatorium-tenants-production.prometheusrules.yaml
+# For local testing:
+# - observatorium-tenants-production.prometheusrules.yaml.test
+
+evaluation_interval: 1m
+
+tests:
+- interval: 1m
+  input_series:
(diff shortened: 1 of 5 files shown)


observability/prometheus_rule_tests/observatorium-tenants.prometheusrulestests.yaml
```diff
@@ -0,0 +1,53 @@
+---
+$schema: /app-interface/prometheus-rule-test-1.yml
+
+rule_files:
+# For upstream:
```
jessicalins, line 5: is there a reason to keep the other files commented out? or can we remove the comments?

jessicalins (commented):
do `observability/prometheus_rule_tests/observatorium-tenants.prometheusrulestests.yaml` and `resources/observability/prometheusrules/observatorium-tenants.prometheusrulestests.yaml` have the same content? I'm not sure why we have them duplicated in 2 files?
Otherwise, as far as I understand, looks good:)

observability/prometheus_rule_tests/observatorium-tenants.prometheusrulestests.yaml
matej-g: This was copied from how other rules are written but it's a good point, I also don't see much value in keeping it here. I will simplify the comment.

matej-g:
> do `observability/prometheus_rule_tests/observatorium-tenants.prometheusrulestests.yaml` and `resources/observability/prometheusrules/observatorium-tenants.prometheusrulestests.yaml` have the same content? I'm not sure why we have them duplicated in 2 files? Otherwise, as far as I understand, looks good:)

I think this is 'by design', as when we generate the rules, the tests are also 'generated' (i.e. copied :sweat_smile:) into the `resources` directory

jessicalins (approved):
:straight_ruler:

PR #106, opened 2021-11-30
Fix configmap filenames

jessicalins:
After https://github.com/rhobs/configuration/pull/104 introduced 'compound' names (e.g. `query_frontend.libsonnet`) and since we were creating the configmap files by splitting the file name by `.`, in this case the generated configmap filename would be `grafana-dashboard-observatorium-thanos-query_frontend.configmap.yaml`, causing an error once it is actually used/deployed:
```bash
Invalid value: "grafana-dashboard-observatorium-thanos-query_frontend". This field must adhere to DNS-1123 subdomain names spec.More info can be found at https://kubernetes.io/docs/concepts/overview/working-with-objects/names/
```
This pr just adds a function that gets the filename and replaces `_` -> `-`. If the filename does not have any `_`, `std.strReplace` ignores it and does not make any changes. 
After this change I just ran `make grafana` and the configmap file was correctly updated.


moadz (approved):
LGTM

matej-g (approved):

PR #110, opened 2021-12-20
Add alerting for PVCs getting filled

matej-g:

This PR adds alerts to improve our visibility into issues with PVC getting filled. It is proposed to add two alerts:
- `ObservatoriumPersistentVolumeUsageHigh` which has low severity and should serve as a warning that the PVC is being filled beyond a level that cannot be sustained long term (i.e. does not require immediate action but needs to be addressed soon).
- `ObservatoriumPersistentVolumeUsageCritical` which has high severity and is meant to alert on cases when the PVC got filled beyond a critical threshold (95%). This requires immediate action, i.e. PVC extension.

Since PVCs are used only on critical components (store, receiver and ruler), I'd suggest to _not_ carve out exclusions for certain namespaces etc. but rather alert on all PVCs.

### DIFF
--- a/docs/sop/observatorium.md
+++ b/docs/sop/observatorium.md
@@ -1136,6 +1136,65 @@ For the set availability guarantees the observatorium api or the thanos receiver
 
 ---
 
+## ObservatoriumPersistentVolumeUsageHigh
+
+### Impact
+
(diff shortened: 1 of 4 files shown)


ianbillett (approved):
LGTM

PR #111, opened 2022-01-10
New alert to check Thanos Store blocks not being loaded

spaparaju:
This PR adds a new RHOBS alert on Thanos Store blocks not being loaded.


ianbillett:
I'm probably missing something here - but I don't understand how this alert satisfies the acceptance criteria of the [ticket](https://issues.redhat.com/browse/MON-1847)? 
> We get an alert if we detect that RHOBS contains less data retained than expected.

spaparaju:
After a data loss incident, initial goal of this [issue](https://issues.redhat.com/browse/MON-1847) was to check for resolutions. After couple of discussions, it was concluded that alerting for missing data is more practical and added a new metric `thanos_bucket_store_blocks_last_loaded_timestamp_seconds`. This PR adds a new alert on this new metric to alert if no block has been uploaded for a duration of 6 hours. 

Will update the [issue](https://issues.redhat.com/browse/MON-1847) with the above context.

docs/sop/observatorium.md
```diff
@@ -1166,6 +1166,40 @@ One or more PVCs are filled to more than 90%. The remaining free space might not
...
+
+## ObservatoriumNoStoreBlocksLoaded
+
+### Impact
+
+Thanos Store blocks are not being loaded. This can indicate possible data loss.
+
+### Summary
+
+During the last 6 hours, not even a single Thanos Store block has been loaded. This can indicate possible data loss.
```
matej-g, line 1179:
```suggestion
During the last 6 hours, not even a single Thanos Store block has been loaded.
```

docs/sop/observatorium.md
```diff
@@ -1166,6 +1166,40 @@ One or more PVCs are filled to more than 90%. The remaining free space might not
...
+
+### Steps
+
+- Check the namespace of RHOBS causing this alert to fire.
+- Locate Thanos Compact Statefulset in the affected namespace.
+- Check these downsampling command line args for Thanos Compact if they are defined as per these guidelines: https://thanos.io/tip/components/compact.md/
+        - --retention.resolution-raw
+        - --retention.resolution-5m
+        - --retention.resolution-1h
+
```
matej-g, line 1202: I'm missing steps on what to do if this is misconfigured - do we need to update it in the AppSRE Interface? Any other aspects we should be looking out for when changing this config?

observability/prometheusrules.jsonnet
```diff
@@ -503,6 +503,20 @@ local renderAlerts(name, environment, mixin) = {
         {
           name: 'observatorium-metrics',
           rules: [
+            {
+              alert: 'ObservatoriumNoStoreBlocksLoaded',
+              annotations: {
+                description: 'Observatorium Thanos Store {{$labels.namespace}}/{{$labels.job}} has not loaded even a single block.',
+                summary: 'Observatorium Thanos Ruler has not loaded even a sigle block. This should not have happened, indicates data loss. Check out the configuration.',
```
matej-g, line 510: Maybe we could add information on how long
```suggestion
                summary: 'Observatorium Thanos Ruler has not loaded a single block in the last 6 hours.',
```

observability/prometheusrules.jsonnet
```diff
@@ -503,6 +503,20 @@ local renderAlerts(name, environment, mixin) = {
...
+              },
+              expr: |||
+                absent(thanos_bucket_store_blocks_last_loaded_timestamp_seconds) != 1 and (time() - thanos_bucket_store_blocks_last_loaded_timestamp_seconds) > 6 * 60 * 60
+              ||| % config.thanos.store,
+              'for': '6h',
+              labels: {
+                severity: 'critical',
```
matej-g, line 517: Should this be `critical`? Should this get someone 'out of the bed' so to speak? :smile:

ianbillett (changes requested):
Sorry @spaparaju but I really don't think that this alert is telling us anything valuable about data loss in our RHOBS infrastructure.

The metric `thanos_bucket_store_blocks_last_loaded_timestamp_seconds` tells us the most recent time that `Store` loaded a block from object storage.

If this exceeds 6 hours, this does not indicate data loss (i.e. deleting blocks from object storage), this would indicate that new blocks are not being uploaded to object storage.

Blocks are uploaded via `Receive` or `Compactor`, for which we already have fairly comprehensive set of alerts - which leads me to think that we already have the 'alerting on data not being uploaded' path covered.

spaparaju:
@ianbillett the metric  `thanos_bucket_store_blocks_last_loaded_timestamp_seconds` is updated every time a new block is found (repeated by  `--sync-block-duration` flag of the Store) to be synced between local (disk) and remote view. 

Last time when we ran into data loss issue, based on Compactor flags we chose resulted in the data loss i.e. Compactor did not create any new bigger blocks as expected.  This alert tries to address this specific path which might result in the data loss.

Question: From `Receive` / `Compactor`,  can you please point me to any existing alerts that we can leverage for the above scenario which has resulted in data loss ?

matej-g:
I think we might just all understand the 'data loss' differently here, if I'm reading the conversations correctly. While this alert does not mean we lost existing data, it is indicating that even though we should be continuously seeing new data being loaded, there was nothing loaded in last 6 hours - this can indicate that there is something wrong with data ingestion and that we are 'losing' this data, since we're not seeing them being loaded and they perhaps are getting 'lost' somewhere on the way to us (e.g. due to misconfiguration). 

Does that make sense @spaparaju @ianbillett? 

As for similar alerts, we are alerting on compactor not uploading in the last 24 hours and there is also `ThanosReceiveNoUpload`, but I guess we don't have similar alerting from the store side.

docs/sop/observatorium.md
```diff
@@ -1166,6 +1166,40 @@ One or more PVCs are filled to more than 90%. The remaining free space might not
...
+  - `observatorium-mst-production`
+
+### Steps
+
+- Check the namespace of RHOBS causing this alert to fire.
+- Locate Thanos Compact Statefulset in the affected namespace.
+- Check these downsampling command line args for Thanos Compact if they are defined as per these guidelines: https://thanos.io/tip/components/compact.md/
+        - --retention.resolution-raw
+        - --retention.resolution-5m
+        - --retention.resolution-1h
```
ianbillett, line 1201: This alert would not have fired in the compactor incident as `Receive` was still uploading blocks to object storage that `Store` was syncing and serving.

docs/sop/observatorium.md
spaparaju: Currently setting all the resolutions (raw, 5m, 1h) to the same duration makes Compactor works as expected. I dont think this fact is not documented at  https://thanos.io/tip/components/compact.md/.  There are other min. values for 5m & 1h resolutions which Thanos currently does not verify for these min. values for the respective Cmd line args.

spaparaju:
> I think we might just all understand the 'data loss' differently here, if I'm reading the conversations correctly. While this alert does not mean we lost existing data, it is indicating that even though we should be continuously seeing new data being loaded, there was nothing loaded in last 6 hours - this can indicate that there is something wrong with data ingestion and that we are 'losing' this data, since we're not seeing them being loaded and they perhaps are getting 'lost' somewhere on the way to us (e.g. due to misconfiguration).
> 
> Does that make sense @spaparaju @ianbillett?
Yes @matej-g . This alert is on the Store component.  When RHOBS does experience a data loss i.e users not able to see any data (there can be n-possible reasons, one reason resulted in the outage that we experienced is because down-sampled-resolutions are not same and min. values are not currently enforced at the Thanos Compact), the Store component would not be loading any blocks from the Object store (Thanos Store syncs blocks as per `--sync-block-duration`). The metric `thanos_bucket_store_blocks_last_loaded_timestamp_seconds` captures the timestamp whenever sync. can load the blocks and this alert is based on if no block loads happen for the last 6 hours.
> 
> As for similar alerts, we are alerting on compactor not uploading in the last 24 hours and there is also `ThanosReceiveNoUpload`, but I guess we don't have similar alerting from the store side.

docs/sop/observatorium.md
spaparaju: When RHOBS does experience a data loss i.e users not able to see any data (there can be n-possible reasons, one reason resulted in the outage that we experienced is because down-sampled-resolutions are not same and min. values are not currently enforced at the Thanos Compact), the Store component would not be loading any blocks from the Object store (Thanos Store syncs blocks as per `--sync-block-duration`). The metric `thanos_bucket_store_blocks_last_loaded_timestamp_seconds` captures the timestamp whenever sync. can load the blocks and this alert is based on if no block loads happen for the last 6 hours.

observability/prometheusrules.jsonnet
spaparaju: Changed to `High`.  As per my experience with SRE while adding new alerts, SRE  would have an opinion as well. :)

matej-g:
Thanks for the explanation @spaparaju :+1:, do we have now common understanding @ianbillett? Do you still feel like this alert is lacking merit?

docs/sop/observatorium.md
```diff
@@ -1166,6 +1166,40 @@ One or more PVCs are filled to more than 90%. The remaining free space might not
...
+  - `observatorium-mst-production`
+
+### Steps
+
+- Check the namespace of RHOBS causing this alert to fire.
+- Locate Thanos Compact Statefulset in the affected namespace.
+- Currently Thanos Compact works as expected if the retention.resolution-raw, retention.resolution-5m and retention.resolution-1h are set for the same duration. 
+- Also check guidelines for these downsampling Thanos Compact command line args at: https://thanos.io/tip/components/compact.md/
+        - --retention.resolution-5m needs more than 40 hours
+        - --retention.resolution-1h needs to be more than 10 days
```
ianbillett, line 1201: For us to accept this alert, we need to add more context to the steps outlined here to help the engineer responding to the task. 

* This alert tells us that `Store` has not synced any blocks in the past 6h. This could be due to multiple reasons:
  * Thanos Store is unable to communicate with object store.
  * No new blocks are being uploaded to object store. This indicates that both `Receive` and `Compact` are not uploading new blocks.
* Check on the health of the affected `Store` component.
  * Check the logs - are there any errors?
  * Check the `Store` dashboard - are there any obvious problems?
  * Check the secrets `Store` is using to communicate with Object storage.
 * Validate that both `Receive` and `Compact` are uploading new blocks to Object storage.
   * Investigate Receive
     * Check the `Receive` logs.
     * Check the `Receive` dashboard.
   * Investigate Compactor
     * Check the `Compactor` logs.
     * Check the `Compactor` dashboard.
     * Check the `Compactor` configuration:
       * `<insert the compactor config here>`
    * Investigate Object Storage (you will need AppSRE's help)
      * When was the last time that a block was uploaded?

ianbillett (commented):
There is some value in this alert. However, if we are going to accept this alert, we need to be much clearer about what this alert means, what the potential causes are and what the debugging steps a responder would take. The compactor configuration options is only a small part of this story.

observability/prometheusrules.jsonnet
```diff
@@ -503,6 +503,20 @@ local renderAlerts(name, environment, mixin) = {
         {
           name: 'observatorium-metrics',
           rules: [
+            {
+              alert: 'ObservatoriumNoStoreBlocksLoaded',
+              annotations: {
+                description: 'Observatorium Thanos Store {{$labels.namespace}}/{{$labels.job}} has not loaded even a single block.',
+                summary: 'Observatorium Thanos Store has not loaded even a sigle block in the last 6 hours.',
```
ianbillett, line 510:
```suggestion
                description: 'Observatorium Thanos Store {{$labels.namespace}}/{{$labels.job}} has not loaded any blocks in the last 6 hours.',
                summary: 'Observatorium Thanos Store has not loaded any blocks in the last 6 hours.',
```

resources/observability/prometheusrules/observatorium-thanos-production.prometheusrules.yaml
```diff
@@ -362,23 +362,6 @@ spec:
       labels:
         service: telemeter
         severity: high
-    - alert: ThanosReceiveTrafficBelowThreshold
```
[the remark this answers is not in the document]
spaparaju: it was caused by Jsonnet dependency resolution. Fixed this problem, not changing anything related to `ThanosReceiveTrafficBelowThreshold` alert

docs/sop/observatorium.md
spaparaju: Great suggestions. Updated context for this alert.

matej-g (approved):
Looking good now :+1: I have one more optional suggestion

matej-g:
@ianbillett satisfied with the state of the alert or are there still any unresolved issues for you?

ianbillett (approved):
LGTM - Let's merge this puppy :muscle:

matej-g:
Shall we roll out the alert as well? We will probably need to bump Thanos to the latest (IIRC `v0.24.0`) in our environments cc @spaparaju

PR #114, opened 2022-01-14
Add SLO alerts for Telemeter

ianbillett:
This PR adds alerts & rules for SLO violations using the existing slo-libsonnet library - [MON-2153](https://issues.redhat.com/browse/MON-2153).

The library generates a large number of new rules & alerts, PR reviewers are best advised to primarily review the jsonnet changes.

We'll add alerts for MST once we're happy with this set of alerts & rules for telemeter in staging.


observability/prometheusrules.jsonnet
```diff
@@ -228,30 +228,113 @@ local renderAlerts(name, environment, mixin) = {
...
+          latencyTarget: 10.761264004567169,
+          latencyBudget: 0.9,
+        }),
+        slo.latencyburn({
+          alertName: 'TelemeterAPIMetricsReadLatencyErrorBudgetBurning',
+          alertMessage: 'API /query endpoint is burning too much error budget to gurantee latency SLOs',
+          metric: 'up_custom_query_duration_seconds_bucket',
+          // We can't use !~ operator in these selectors
+          selectors: ['query="query-path-sli-100M-samples"'],
+          latencyTarget: 21.6447457021712,
```
moadz, line 336: This is going to alert constantly rn right?

moadz (approved):
LGTM but I the latency SLO's will alert spam if this is merged currently.

observability/prometheusrules.jsonnet
```diff
@@ -228,30 +228,113 @@ local renderAlerts(name, environment, mixin) = {
 };
 
 {
+  // This set of SLOs are driven by the RHOBS Service Level Objectives document
```
matej-g, line 231: :clap:

observability/prometheusrules.jsonnet
```diff
@@ -228,30 +228,113 @@ local renderAlerts(name, environment, mixin) = {
...
+          // We can't use !~ operator in these selectors
+          selectors: ['job="telemeter-server"', 'handler=~"upload|receive"', 'code=~"^(2..|3..|5..)$"'],
+          latencyTarget: 5,
+          latencyBudget: 0.9,
         }),
       ],
     },
     {
-      name: 'telemeter-authorize.slo',
+      name: 'telemeter-api-metrics-write-availability.slo',
```
matej-g, line 262: I find these names a little bit confusing, since I assume we're talking about Observatorium API here and Telemeter part of the name refers to the instance (in previous alerts, we talk about 'Telemeter server'). At risk of it being too long of a name, should we still throw in 'Observatorium' before 'API'?

observability/prometheusrules.jsonnet
```diff
@@ -228,30 +228,113 @@ local renderAlerts(name, environment, mixin) = {
...
       slos: [
         slo.errorburn({
-          alertName: 'TelemeterAuthorizeErrorBudgetBurning',
-          alertMessage: 'Telemeter /authorize is burning too much error budget to gurantee overall availability',
-          metric: 'haproxy_server_http_responses_total',
-          selectors: ['route="telemeter-server-authorize"'],
-          errorSelectors: ['code="5xx"'],
-          target: 0.98,
+          alertName: 'TelemeterAPIMetricsWriteAvailabilityErrorBudgetBurning',
+          alertMessage: 'API /receive handler is burning too much error budget to gurantee availability SLOs',
```
matej-g, line 266: nit: Looks like we use interchangeably handler and endpoint in the messages, would be nice to unify on one term

matej-g (approved):
Looking good :tada:  few minor nits + also curious about @moadz question

observability/prometheusrules.jsonnet
ianbillett: Yep that's right - I will add in relevant silences before rolling this out into out environments.

observability/prometheusrules.jsonnet
ianbillett: Good spot - will get that updated.

observability/prometheusrules.jsonnet
ianbillett: Yeah really nice point - I was thinking something similar. We should have a loose standard that we follow...

`rhobs-<instance>-<component>-<slo>` or something?

Full disclosure - I left `telemeter` at the front to avoid some jsonnet weirdness / wrangling about dashboard names.

matej-g: I had in mind exactly something like that :+1:

ianbillett:
Thanks both for the helpful comments :+1:

PR #117, opened 2022-01-24
Remove query label from alerts

ianbillett:
SLO alerts failing validation [here](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/31686)

### DIFF
--- a/observability/prometheusrules.jsonnet
+++ b/observability/prometheusrules.jsonnet
@@ -122,6 +122,7 @@ local appSREOverwrites(environment) = {
     labels: std.prune(labels {
       group: null,
       code: null,
+      query: null,
     }),
   },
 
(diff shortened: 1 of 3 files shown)


matej-g (approved):

observability/prometheusrules.jsonnet
```diff
@@ -122,6 +122,7 @@ local appSREOverwrites(environment) = {
     labels: std.prune(labels {
       group: null,
       code: null,
+      query: null,
```
jessicalins, line 125: I don't really get why removing `query` fixes https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/107177/artifact/temp/reports/index.html - maybe we can go through that together? From what I understood it just mentions "Validation error" but it does not say exactly where is failing? Could be I'm also missing something..

ianbillett: I'll show ya :)

PR #118, opened 2022-01-25
Add MST SLO alerts & refactor SLO generator

ianbillett:
This PR adds SLO alerts for the MST instance. Given the length of the app-interface MR cycles it makes sense to define them all now and only go through one more iteration cycle.

We also refactor out the common parts from both the Telemeter & MST SLOs.

We will only ship the production alerts once we've checked staging & when all of the associated run books have been written.

### DIFF
--- a/observability/prometheusrules.jsonnet
+++ b/observability/prometheusrules.jsonnet
@@ -89,6 +89,10 @@ local appSREOverwrites(environment) = {
       else if
         std.startsWith(name, 'rhobs-telemeter') && environment == 'stage' then '080e53f245a15445bdf777ae0e66945d'
       else if
+        std.startsWith(name, 'rhobs-mst') && environment == 'production' then '283e7002d85c08126681241df2fdb22b'
+      else if
+        std.startsWith(name, 'rhobs-mst') && environment == 'stage' then '92520ea4d6976f30d1618164e186ef9b'
+      else if
         std.startsWith(name, 'gubernator') then 'no-dashboard'
       else error 'no dashboard id for group %s' % name,
   },
@@ -234,9 +238,27 @@ local renderAlerts(name, environment, mixin) = {
 };
 
 {
+  local flatten(originalSLOs) = {
+    prometheusAlerts+:: {
+      groups: [
+        {
+          local slos = [
+            slo.alerts + slo.recordingrules
+            for slo in group.slos
+          ],
+
+          name: group.name,
+          rules: std.flattenArrays(slos),
+        }
+        for group in originalSLOs
+      ],
+    },
+  },
+
   // This set of SLOs are driven by the RHOBS Service Level Objectives document
   // https://docs.google.com/document/d/1wJjcpgg-r8rlnOtRiqWGv0zwr1MB6WwkQED1XDWXVQs/edit
-  local telemeterSLOs = [
+
+  local telemeterServerSLOs = [
     {
       name: 'rhobs-telemeter-telemeter-server-metrics-write-availability.slo',
       slos: [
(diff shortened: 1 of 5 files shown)


observability/prometheusrules.jsonnet
```diff
@@ -264,105 +288,104 @@ local renderAlerts(name, environment, mixin) = {
...
-    },
+  local apiSLOs = function(instance, upNamespace, apiJob) {
+    local apiJobSelector = 'job="' + apiJob + '"',
+    local upNamespaceSelector = 'namespace="' + upNamespace + '"',
+    slos: [
+      {
+        name: 'rhobs-' + instance + '-api-metrics-write-availability.slo',
+        slos: [
+          slo.errorburn({
+            alertName: 'APIMetricsWriteAvailabilityErrorBudgetBurning',
```
ianbillett, line 301: Not appending the instance name to the alert name is deliberate - this way we will re-use the run books across each instance.

observability/prometheusrules.jsonnet
```diff
@@ -264,105 +286,104 @@ local renderAlerts(name, environment, mixin) = {
...
+            alertMessage: 'API /receive handler is burning too much error budget to gurantee availability SLOs',
```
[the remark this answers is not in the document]
jessicalins: same for the other `alertMessage`s below

observability/prometheusrules.jsonnet
```diff
@@ -264,105 +286,104 @@ local renderAlerts(name, environment, mixin) = {
...
+            errorSelectors: ['code=~"5.+"'],
+            target: 0.95,
+          }),
+        ],
+      },
+      {
+        name: 'rhobs-' + instance + '-api-metrics-write-latency.slo',
+        slos: [
+          slo.latencyburn({
+            alertName: 'TelemeterAPIMetricsWriteLatencyErrorBudgetBurning',
```
jessicalins, line 314: I think here we want to drop the instance name too?
```suggestion
            alertName: 'APIMetricsWriteLatencyErrorBudgetBurning',
```

observability/prometheusrules.jsonnet
```diff
@@ -264,105 +286,104 @@ local renderAlerts(name, environment, mixin) = {
...
+          }),
+        ],
+      },
+      {
+        name: 'rhobs-' + instance + '-api-metrics-read-latency.slo',
+        slos: [
+          slo.latencyburn({
+            alertName: 'APIMetricsReadLatencyErrorBudgetBurning',
+            alertMessage: 'API /query endpoint is burning too much error budget to gurantee latency SLOs',
+            metric: 'up_custom_query_duration_seconds',
```
jessicalins, line 351: was `_bucket` dropped on purpose? before was `up_custom_query_duration_seconds_bucket` (same for the ones below)

jessicalins (commented):
nice refactor! :parrot: just left some questions/suggestions

observability/prometheusrules.jsonnet
ianbillett: Because the `slo-libsonnet` library implicitly adds the `_bucket` and `_count` prefixes to these metrics :) Before we had `..._bucket_bucket` :sweat_smile:

observability/prometheusrules.jsonnet
ianbillett: Yep - that is visible in the job alert [label](https://github.com/rhobs/configuration/pull/118/files#diff-ddada270f4b62e535a6e35182a15f088d98dbfbd51c64bc2aae7696c38bf7bddR59).

observability/prometheusrules.jsonnet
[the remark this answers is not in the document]
ianbillett: Woops! Fixing.

observability/prometheusrules.jsonnet
ianbillett: Yes! Good spot.

observability/prometheusrules.jsonnet
jessicalins: ah right! makes sense :+1:

jessicalins (approved):
:slot_machine:

matej-g (approved):

PR #121, opened 2022-01-27
Updating Thanos QFE to use shared memcache instead of in-memory cache

moadz:
Currently, all Thanos QFE replicas have their own cache. This change adds a shared memcache instance (with some placeholder pod sizing) to share the cache between the replicas.


ianbillett (approved):
LGTM - let's get this puppy into staging :muscle:

matej-g (approved):
Niceee, one optional comment, happy to merge now

services/observatorium-metrics.libsonnet
```diff
@@ -532,7 +538,55 @@ local tenants = (import '../configuration/observatorium/tenants.libsonnet');
     }),
 
     // For now, just use an in-memory cache.
-    queryFrontendCache:: {},
+    queryFrontendCache:: memcached({
```
jessicalins, line 541: I think we can update the comment above to reflect this change?

services/observatorium-metrics.libsonnet
```diff
@@ -506,11 +506,17 @@ local tenants = (import '../configuration/observatorium/tenants.libsonnet');
       ],
       serviceMonitor: true,
       queryRangeCache: {
-        type: 'in-memory',
-        config+: {
-          max_size: '0',
-          max_size_items: 2048,
-          validity: '6h',
+        type: 'memcached',
```
jessicalins, line 509: I wonder how did you get to those values? mostly curious about the approach (like giving it more room for now and after seeing how it behaves in production, tweak the values, etc..)

jessicalins: as I understand now we switch to a shared memcache - could it be that the shared memcache gets overloaded at some point? or ideally via this configuration we already give enough room for that?
in case memcache goes down for some reason, how much do we care? (main consequence would be cache invalidation?)

jessicalins (commented):
looks good! :shaved_ice: I left some questions for my own understanding and to try getting more context about the change :)

PR #123, opened 2022-01-27
Increasing index cache memcache get multi concurrency

moadz:
Currently we're suffering from intermittent `ThoreStoreGrpcErrorRate` increases that we have attributed to `get_multi` timeouts between Thanos store shards and their respective memcache instances. 

### Around the time Thanos Store started timing out we saw an increase in `getmulti` timeouts on the index-cache
image

### Looking further we see an equivalent spike in `thanos_memcached_getmulti_gate_duration_seconds_bucket` (requests waiting to be dispatched to memcache) 
image

### We also see Thanos store executing the `max_get_multi` requests it is configured to (1000 prior to this PR) 
image

We would like to uncap this limit to stop get_multi requests timing out waiting for their turn. 

[Queries in question](https://prometheus.telemeter-prod-01.devshift.net/graph?g0.range_input=1h&g0.end_input=2022-01-27%2016%3A00&g0.expr=thanos_memcached_getmulti_gate_queries_in_flight%7Bnamespace%3D%22observatorium-metrics-production%22%2C%20name%3D%22index-cache%22%7D&g0.tab=0&g1.range_input=1h&g1.end_input=2022-01-27%2016%3A00&g1.expr=histogram_quantile(0.90%2C%20rate(thanos_memcached_getmulti_gate_duration_seconds_bucket%7B%20namespace%3D%22observatorium-metrics-production%22%2C%20name%3D%22index-cache%22%7D%5B5m%5D))&g1.tab=0&g2.range_input=1h&g2.end_input=2022-01-27%2016%3A00&g2.expr=rate(thanos_memcached_operation_failures_total%7Bnamespace%3D%22observatorium-metrics-production%22%2C%20operation%3D%22getmulti%22%2C%20name%3D%22index-cache%22%2C%20reason%3D%22timeout%22%7D%5B5m%5D)&g2.tab=0)

### DIFF
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -2181,7 +2181,7 @@ objects:
               "max_async_buffer_size": 200000
               "max_async_concurrency": 200
               "max_get_multi_batch_size": 100
-              "max_get_multi_concurrency": 1000
+              "max_get_multi_concurrency": 0
               "max_idle_connections": 1300
               "max_item_size": "5MiB"
               "timeout": "2s"
@@ -2427,7 +2427,7 @@ objects:
               "max_async_buffer_size": 200000
               "max_async_concurrency": 200
               "max_get_multi_batch_size": 100
-              "max_get_multi_concurrency": 1000
+              "max_get_multi_concurrency": 0
               "max_idle_connections": 1300
               "max_item_size": "5MiB"
               "timeout": "2s"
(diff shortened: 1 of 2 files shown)


ianbillett:
Any reason we're not setting it to some much higher level? i.e. 10k?

matej-g:
> Any reason we're not setting it to some much higher level? i.e. 10k?

I guess the idea was to go unlimited and see how much we can get away with in staging and then maybe base the value on that. I'm also not sure if e.g. memcached will start to give out at 3k if having 10k would get us any benefit.

moadz:
> Any reason we're not setting it to some much higher level? i.e. 10k? The limit it there for some reason?
Argument is, we don't know the peak we might observe, but @matej-g has been able to reproduce the specific query that causes this. So we will uncap in staging, throw the query at it and see what the peak # of getMulti's is + if there are any memcache issues we run into with a concurrency limit that high. 

We already have connection limits on the memcache side, this is just the client side limit in thanos-store.

ianbillett (approved):

PR #127, opened 2022-01-28
Increasing getMultiBatchSize on indexCache memcached instance

moadz:
Upon removing concurrency limits we're still observing timeouts on requests to memcache from store gateway. 

* Memached cpu/memory utilization remains minimal, so we believe there is capacity to service one large getMulti requests instead of several smaller ones
* StoreGW appears to be operating near its CPU limit, which might explain the timeouts on requests with no CPU resources to fulfil them. Reducing the number of requests may help with this also

### DIFF
--- a/services/observatorium-metrics.libsonnet
+++ b/services/observatorium-metrics.libsonnet
@@ -246,7 +246,7 @@ local tenants = (import '../configuration/observatorium/tenants.libsonnet');
           timeout: '2s',  // default: 500ms
           max_async_buffer_size: 200000,  // default: 10_000
           max_async_concurrency: 200,  // default: 20
-          max_get_multi_batch_size: 100,  // default: 0 - No batching.
+          max_get_multi_batch_size: 0,  // default: 0 - Batch every request into one.
           max_get_multi_concurrency: 0,  // default: 100
           max_item_size: '5MiB',  // default: 1Mb
         },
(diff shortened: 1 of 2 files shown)


matej-g (approved):
:zero:

services/observatorium-metrics.libsonnet
```diff
@@ -246,7 +246,7 @@ local tenants = (import '../configuration/observatorium/tenants.libsonnet');
           timeout: '2s',  // default: 500ms
           max_async_buffer_size: 200000,  // default: 10_000
           max_async_concurrency: 200,  // default: 20
-          max_get_multi_batch_size: 100,  // default: 0 - No batching.
+          max_get_multi_batch_size: 0,  // default: 0 - Batch every request into one.
```
ianbillett, line 249: The comment above makes me think that `0` means that there is no batching going on?

moadz: The comment is incorrect: 
If `max_get_multi_batch_size` is set to `0`, all key gets are matched into a single `getMulti` (which is a batched request). So i think it's a bit confusing as the flag denotes the batching of the batched requests. I'll revert to the old comment to simplify
https://github.com/thanos-io/thanos/blob/main/pkg/cacheutil/memcached_client.go#L444-L451

matej-g: It's a bit misleading - yeah, no batching is going on, which means instead of splitting the request into smaller batches, we 'batch' it into a single request (which we want here). So, yeah, basically we're _not_ batching but making one bigger request. Our finding was that it seemed we are perhaps batching too much, meaning there are a too many of 'small' requests fired as opposed to fewer 'bigger' requests.

PR #130, opened 2022-01-31
Enable Rules API in Observatorium

onprem:
This PR
- Adds Thanos Rule Syncer sidecar to Thanos Rulers
- Deploys `rules-objstore` as Rules Storage Backend
- Enables Rules API in `observatorium-api`

The secret with object storage config for Rules Backend will be created in AppSRE vault, and the secret with S3 credentials will be created via app-interface.


resources/services/observatorium-metrics-template.yaml
```diff
@@ -1593,6 +1593,7 @@ objects:
           - --tsdb.retention=48h
           - --tsdb.block-duration=2h
           - --query=dnssrv+_http._tcp.observatorium-thanos-query.${NAMESPACE}.svc.cluster.local
+          - --rule-file=/etc/thanos/rules/observatorium-rule-syncer.yaml
```
ianbillett, line 1596: What are the semantics for updates? They just happen auto-magically when the file is updated?

resources/services/observatorium-template.yaml
```diff
@@ -677,6 +679,117 @@ objects:
...
+            failureThreshold: 10
+            httpGet:
+              path: /live
+              port: 8081
+              scheme: HTTP
+            periodSeconds: 30
+          name: rules-objstore
+          ports:
+          - containerPort: 8081
+            name: internal
```
ianbillett, line 742: Do we need to add a service monitor or something to make sure we get metrics in Prometheus?

ianbillett (approved):
Couple of nits - but nothing blocking, let's get this puppy into staging!

resources/services/observatorium-template.yaml
onprem: The jsonnet lib supports adding a ServiceMonitor, but `rules-objstore` is not instrumented with metrics yet. Once we have some instrumentation there, I'll enable the service monitor in a follow up PR.

onprem: Oh I think we expose the default Go metrics already, so I'll enable the service monitor in this PR itself. Created observatorium/rules-objstore#8 to track adding custom instrumentation.

PR #137, opened 2022-02-01
Add Observatorium Alertmanager for Tenant Alerting

ianbillett:
This PR implements the basic skeleton of our Tenant Alerting capabilities in RHOBS.

This was tested and working as expected in a temporary cluster I got from cluster bot.

Tenant alertmanager configuration is defined, templated and deployed in app-interface - see [this MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/32713).

There are a number of bits missing from this that we can add in due course - I'm just keen to get _something_ out there that we can start testing with / iterating from.


resources/services/observatorium-metrics-template.yaml
```diff
@@ -1594,6 +1714,7 @@ objects:
           - --tsdb.block-duration=2h
           - --query=dnssrv+_http._tcp.observatorium-thanos-query.${NAMESPACE}.svc.cluster.local
           - --rule-file=/etc/thanos/rules/observatorium-rule-syncer.yaml
+          - --alertmanagers.url=observatorium-alertmanager.${NAMESPACE}.svc.cluster.local
```
simonpasquier, line 1717: alerts should be sent to all Alertmanager pods so you want `dnssrv`-based discovery probably. Hint: https://github.com/openshift/cluster-monitoring-operator/blob/190f54e5c127e9375011f6cdf931b67d7be50407/assets/thanos-ruler/alertmanagers-config-secret.yaml#L19-L20

resources/services/observatorium-metrics-template.yaml
```diff
@@ -234,6 +234,126 @@ objects:
...
+  metadata:
+    labels:
+      app.kubernetes.io/name: observatorium-alertmanager
+    name: alertmanager-data
+  spec:
+    accessModes:
+    - ReadWriteOnce
+    resources:
+      requests:
+        storage: 50Gi
```
simonpasquier, line 355: 50Gi is likely to be more than enough

resources/services/observatorium-metrics-template.yaml
```diff
@@ -234,6 +234,126 @@ objects:
...
+- apiVersion: apps/v1
+  kind: StatefulSet
+  metadata:
+    labels:
+      app.kubernetes.io/component: alertmanager
+      app.kubernetes.io/name: alertmanager
+      app.kubernetes.io/part-of: observatorium
+    name: observatorium-alertmanager
+  spec:
+    replicas: 1
```
simonpasquier, line 280: why not 2 replicas?

ianbillett: Purely for simplicity & velocity's sake as we want to get something out there that we can iterate on.

resources/services/observatorium-metrics-template.yaml
ianbillett: What would be a more sensible number?

resources/services/observatorium-metrics-template.yaml
ianbillett: For this case would that then be
```
dnssrv+_web._tcp.observatorium-alertmanager.${NAMESPACE}.svc.cluster.local
```
?

resources/services/observatorium-metrics-template.yaml
simonpasquier: 10Gi should be good

resources/services/observatorium-metrics-template.yaml
simonpasquier: According to https://kubernetes.io/docs/concepts/services-networking/service/#dns, you should target `dnssrv+_ui._tcp.observatorium-alertmanager.${NAMESPACE}.svc.cluster.local`

resources/services/observatorium-metrics-template.yaml
simonpasquier: Ack. Gossip configuration should be straightforward IMHO.
You could look at the Alertmanager statefulset definition from a 4.10 cluster and works backward from there because the deployment should be reasonably hardened (it includes pod disruption budget for instance).

matej-g: I'm fine going forward with having 1 replica but as @simonpasquier it seems not to be too much extra effort, perhaps few more lines in the libsonnet.

matej-g (commented):
Looking good minus comments from @simonpasquier

resources/services/observatorium-metrics-template.yaml
ianbillett: >  few more lines in the libsonnet

matej-g (approved):
Happy to move this forward, I'll leave it up to @ianbillett to follow up with multiple replicas which seem to be the remaining unresolved suggestion.

ianbillett:
I'm going to leave the multi-replica case for now - it gets a bit confusing with the headless services you need to get the individual statefulsets by DNS name and I don't have a particularly fast iteration cycle - created a ticket in the epic to follow up.

PR #143, opened 2022-02-03
Remove storage class from config definition

ianbillett:
image


matej-g (commented):
Shouldn't cluster automatically pick up `gp2` if we do not specify this? cc @moadz

ianbillett:
Good question! Let's try removing it and see what happens.

PR #147, opened 2022-02-04
Adding -vv flag to indexCache memcached config

moadz:
* Why are we manually setting the param in the template? 
a) Args are not parameterised on the memcached observatorium config provider, so adding that would require multiple PR's, a provider update
b) This change is temporary, and will be removed before rolling out into production to get better visibility on where the connections are terminating/timing out 


ianbillett (approved):

matej-g (approved):
-:v::v:

PR #148, opened 2022-02-04
Add OAuth proxy to Observatorium Alertmanager

ianbillett:
I don't know if this is going to work :man_shrugging: But I've followed the prior art set by other services - only one way to find out.

We're adding the Openshift Routes [here](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/32897).

Created the relevant alertmanager-proxy secret [here](https://vault.devshift.net/ui/vault/secrets/app-interface/show/app-sre-stage/telemeter-stage/alertmanager-proxy).


matej-g (approved):
Pending adding the secret(s) to Vault, this is looking fine :+1:

PR #153, opened 2022-02-11
Reintroducing getMulti batching for memcached client 

moadz:
Also removing range query SLO queries as they are too heavyweight

Details in thread: https://coreos.slack.com/archives/CHSQW32Q5/p1644592855006239


matej-g (approved):
:100:

PR #157, opened 2022-02-25
Bug 2058219: ingress nlb metric is not found from telemeter server

tjungblu:
hey there,

QE has found that a metric is missing from telemeter in https://bugzilla.redhat.com/show_bug.cgi?id=2058219

While generating, some new lines appeared that were not related to this change, hope that is alright?


tjungblu:
build seemed to fail on the same issue reported in https://github.com/rhobs/configuration/pull/72#issuecomment-937503124

added another `find -delete` to the respective makefile clean target

matej-g (approved):
Hey @tjungblu thanks for updating the list, looks like the CSI-related metrics have not been added yet, so thanks for including them.

PR #166, opened 2022-03-09
clo and eo metrics for Telemetry

pmoogi-redhat:
This PR adds EO and CLO metrics for Telemetry server allowlist.


pmoogi-redhat (commented):
@moadz kindly review this PR

pmoogi-redhat (commented):
@moadz  kindly review my changes

configuration/telemeter/metrics.json
```diff
@@ -88,6 +102,7 @@
   "{__name__=\"openshift_csi_share_mount_failures_total\"}",
   "{__name__=\"openshift_csi_share_mount_requests_total\"}",
   "{__name__=\"openshift_csi_share_secret\"}",
+  "{__name__=\"pod:eo_es_shards_total:max\"}",
```
moadz, line 105: Is there a reason why we're sending pod-specific metric up to telemetry?

pmoogi-redhat: typically there are 3 instance of es that run as the part of in-cluster logging pipeline. these 3 instances run one per node. no of shards per instance in each one of the three nodes differ based on amount of logs to be indexed. hence max is taken over each instance to characterise how big the index space is becoming within es cluster - formed over three nodes running three instances of es engine.

moadz: SGMT

moadz:
LGTM,  waiting on obs-support to TAL as well

bill3tt:
@pmoogi-redhat I'm slightly concerned about the number of new metrics that you are introducing here - there are 15 new metrics introduce which is a ~13% increase on the number that we already have.

What is the cardinality of each of these metrics within a cluster? I'm trying to get a handle on the impact this is going to have in terms of resources that the telemeter server requires to ingest, store and serve this new data.

configuration/telemeter/metrics.json
jcantrill: @pmoogi-redhat the number of shards for an entire ES cluster should be sufficient for telemetry. We do nothing specific to isolate shards to nodes so they should be reasonably balanced across an ES cluster

configuration/telemeter/metrics.json
```diff
@@ -50,6 +55,11 @@
   "{__name__=\"count:up1\"}",
   "{__name__=\"csv_abnormal\"}",
   "{__name__=\"csv_succeeded\"}",
+  "{__name__=\"eo_es_cluster_management_state_info\"}",
+  "{__name__=\"eo_es_defined_delete_namespaces_total\"}",
+  "{__name__=\"eo_es_misconfigured_memory_resources_info\"}",
```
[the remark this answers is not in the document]
pmoogi-redhat: EO Telemetry metrics list taken from this Ref doc https://docs.google.com/document/d/1ek98QDTztWHuYzUX_T53O9JblUtL9-YRMUK5oHMIK8w/edit 
@periklis and @Red-GV to comment on it.

alanconway:
@bill3tt the info metrics reflect the configuration of the logging singleton, they produce a  single "true" value with a collection of labels indicating the configuration state (features used etc.) Over the lifetime of the cluster this may produce a new timeseries each time the logging configuration changes (expected to be fairly static) but each will only produce a single value so the size of each series is tiny. This is a pattern we tried to follow from elsewhere, shout if you still have concerns.
@pmoogi-redhat I suggest dropping the metrics questioned by @jcantrill , we can re-introduce them later if we find there really is a need. Otherwise LGTM.

configuration/telemeter/metrics.json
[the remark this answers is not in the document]
pmoogi-redhat: pruned further eo metrics - eo_es_defined_delete_namespaces_total,eo_es_misconfigured_memory_resources_info,eo_es_defined_delete_namespaces_total,eo_es_storage_info  removed from our list for Telemetry.

pmoogi-redhat:
@bill3tt  pruned eo metrics list further after core-logging team leads internal discussion. kindly see if this looks fine and added overhead within a given budget per operators (eo + clo)

bill3tt:
@pmoogi-redhat you will need to request that this configuration change is rolled out by `@observatorium-support` in [#forum-observatorium](https://coreos.slack.com/archives/C010A637DGB)

bwplotka:
Please run `make manfiests` to generate required YAMLs (CI error is because of that)

configuration/telemeter/metrics.json
```diff
@@ -72,6 +78,10 @@
   "{__name__=\"job:noobaa_bucket_count:sum\"}",
   "{__name__=\"job:noobaa_total_object_count:sum\"}",
   "{__name__=\"job:noobaa_total_unhealthy_buckets:sum\"}",
+  "{__name__=\"log_forwarder_input_info\"}",
+  "{__name__=\"log_forwarder_output_info\"}",
+  "{__name__=\"log_forwarder_pipeline_info\"}",
+  "{__name__=\"log_logging_info\"}",
```
bwplotka, line 84: This is interesting, are we should it will not introduce big cardinality problems? What are the labels on this metric?

bwplotka:
> ere - there are 15 new metrics introduce which is a ~13% increase on the number that we already have.
 
There are much more series behind those metric names BTW @bill3tt

bwplotka:
@alanconway do you have an example series to validate? If those labels change over time very often and are unique across cluster, we might have a problem

bwplotka:
All those `info` metrics are nice for debugging purpose, but as time series, they tend to be highly cardinal (e.g commit hash that was used to build a certain binary 😱 ).

configuration/telemeter/metrics.json
pmoogi-redhat: @bwplotka these metrics help finding usage pattern of different input and output configurations of logging pipeline. For a given client-cluster the configurable input source and output endpoint may not change over time. Here is the reference doc that is used for implementing and exporting these metrics https://docs.google.com/document/d/1s7odpnziZSaAsnFFHgsZs12dFzgXcy9eNJOSO4l1qCI/edit#heading=h.9zhbu8l4oepw.

alanconway:
@bwplotka Are there any existing services that gather this kind of data - versions deployed, features used etc?
If so are there best practices we can follow? This pattern is one I found while googling but if it's a bad pattern we should replace it with a good one.

pmoogi-redhat:
@bwplotka @bill3tt  we have followed other info type metrics here.

pmoogi-redhat:
@bwplotka @bill3tt as my changes I see now approved. I have also resolved merge conflicts. Kindly let me know if this PR can be merged. CC @alanconway @jcantrill

PR #169, opened 2022-03-21
Add parallel stateless Ruler

saswatamcode:
This PR,

- Upgrades `kube-thanos` to latest
- Adds stateless Ruler StatefulSet alongside the existing one (uses the same ConfigMap for rules and adds one for remote write config)
- Adds new ruler to dashboard queries

Open question: Should we configure all the same rules on both Rulers, as this PR does? Or try this out with one rule first?
cc: @matej-g

### DIFF
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -2135,7 +2135,6 @@ objects:
           - --tsdb.retention=48h
           - --tsdb.block-duration=2h
           - --query=dnssrv+_http._tcp.observatorium-thanos-query.${NAMESPACE}.svc.cluster.local
-          - --rule-file=/etc/thanos/rules/rule-syncer/observatorium.yaml
           - --alertmanagers.url=dnssrv+http://observatorium-alertmanager.${NAMESPACE}.svc.cluster.local:9093
           - |-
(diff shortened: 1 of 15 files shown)


resources/services/observatorium-metrics-template.yaml
```diff
@@ -1772,7 +2137,6 @@ objects:
           - --query=dnssrv+_http._tcp.observatorium-thanos-query.${NAMESPACE}.svc.cluster.local
           - --rule-file=/etc/thanos/rules/rule-syncer/observatorium.yaml
```
matej-g, line 2138: Should we remove all `rule-file` parameters from stateful rulers?

saswatamcode: I think this rule-file parameter is only for rules configured via Observatorium Rules API. I don't think we have any rules configured that way so maybe we don't we need to?

matej-g: I thought we have already introduced some rules via API, am I mistaken? Either way, if anyone adds rules with this change, we would be having both sets of rulers evaluating, right?

matej-g (approved):
Thanks you Saswata :bow: This is looking good to me now.

PR #172, opened 2022-03-28
Update AM labels

jessicalins:
This PR updates the Alertmanager service to contain the (I think) necessary labels so that the service monitor can discover Alertmanager properly. ([Slack thread](https://coreos.slack.com/archives/CCRND57FW/p1648452021640399) for context, [ticket](https://issues.redhat.com/browse/MON-2280))

After running `make manifests`, `imagePullPolicy: IfNotPresent` line was also added - not sure if we want this though, since it was not there before.


saswatamcode:
`oc` seems to panic, but can't replicate that locally.

jessicalins:
@saswatamcode same... it happened in the last 2 commits to main too: https://github.com/rhobs/configuration/commits/main

saswatamcode:
Yeah was also noticed here: https://github.com/rhobs/configuration/pull/171#issuecomment-1080367350

PhilipGough:
> After running make manifests, imagePullPolicy: IfNotPresent line was also added - not sure if we want this though, since it was not there before.

Looks like it is being pulled in from https://github.com/thanos-io/kube-thanos/pull/261 and probably we need to override and set to `Always` (the default)?

jessicalins:
> > After running make manifests, imagePullPolicy: IfNotPresent line was also added - not sure if we want this though, since it was not there before.
> 
> Looks like it is being pulled in from [thanos-io/kube-thanos#261](https://github.com/thanos-io/kube-thanos/pull/261) and probably we need to override and set to `Always` (the default)?

hm good point..we already use `imagePullPolicy` in other places too (also some related with memcached) https://github.com/rhobs/configuration/search?q=imagePullPolicy
Using `IfNotPresent`

PhilipGough:
> we already use imagePullPolicy using `IfNotPresent`

That might be totally ok, particularly if we are referencing specific tags/manifests. Just calling it out that the diff here moves away from the default `Always` which we may or may not want to do :)

ianbillett (approved):
LGTM

jessicalins:
FYI after updating with main + the `imagePullPolicy` changes were updated as well - so now we have just the AM labels change. I've ran also manually/locally `make manifests` to make sure and looks fine.

PhilipGough (approved):

PR #175, opened 2022-03-29
Make stateless ruler pod name less than 63 char

saswatamcode:
This MR shortens the name of metric federation stateless Ruler StatefulSet to `observatorium-metrics-fed-stateless-ruler`, as we face errors when deploying,
```
create Pod observatorium-thanos-metric-federation-stateless-rule-0 in StatefulSet observatorium-thanos-metric-federation-stateless-rule failed error: Pod "observatorium-thanos-metric-federation-stateless-rule-0" is invalid: metadata.labels: Invalid value: "observatorium-thanos-metric-federation-stateless-rule-68fd8f6476": must be no more than 63 characters
```


matej-g (approved):
You're a :star: @saswatamcode

PR #183, opened 2022-04-04
Bump rules-objstore version and add labels

saswatamcode:
This PR bumps the rules-objstore version to the latest to get the newly added metrics in https://github.com/observatorium/rules-objstore/pull/9.

Exisiting rules-objstore was also instrumented with Go metrics and object store Thanos metrics. But it isn't listed in Prometheus AppSRE targets, so adding labels too, to see if that fixes it.


jessicalins (approved):
LGTM :)

PR #185, opened 2022-04-04
Add rhobs-mst role binding

jessicalins:
* Adds `service-account-observatorium-rhobs-mst-staging` and `service-account-observatorium-rhobs-mst`, so that the RHOBS tenant in MST can have proper access
* I've used the same `rhobs-write` and `rhobs-read` roles, since the tenant is the same (`rhobs`)


PhilipGough (approved):
/lgtm

PR #186, opened 2022-04-05
Removing rules-objstore temporarily

saswatamcode:
This PR removes the `rules-objstore` deployment temporarily.

The selectors for `rules-objstore` Deployment were changed in https://github.com/rhobs/configuration/pull/183. As it is an immutable field as per [docs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates), the Deployment would need to be deleted and then redeployed. 

As this is not automatic yet in app-interface, this causes errors in the pipeline too. So the solution is to remove it here to delete the deployment and then revert this commit and promote it to redeploy. :)

For context: [slack thread](https://coreos.slack.com/archives/CCRND57FW/p1649082382403189), [pipeline run](https://console-openshift-console.apps.app-sre-prod-01.i7w5.p1.openshiftapps.com/k8s/ns/observatorium-pipelines/tekton.dev~v1beta1~PipelineRun/saas-observatorium-mst-osd-stage-202204041227/logs/openshift-saas-deploy)

cc: @jessicalins


jessicalins (approved):
:wave:

PR #201, opened 2022-04-20
Update SLO alerts with revised query latency targets

bill3tt:
Like https://github.com/rhobs/configuration/pull/200, but for our SLO alerts.


moadz (approved):
LGTM

PR #202, opened 2022-04-21
Add new tenants to Ruler remote-write config

saswatamcode:
This PR adds the Ruler remote write config for the newly added tenant i.e, rhacs, cnvqe, psiocp and rhods.
Without this, tenants can write Rules but not query them.


moadz (approved):
LGTM but we should probably think about how we want to onboard tenants onto this feature (default on for all or selective onboarding)

PR #207, opened 2022-04-26
Add queue config to ruler remote-write

saswatamcode:
This PR adds queue config to the Ruler remote write configs. It is separate for each tenant in jsonnet as we may want to customize it for each tenant in the future. Default config taken from https://thanos.io/tip/components/rule.md/


moadz (approved):
LGTM

PR #184, opened 2022-04-04
Remove stateful Rulers

saswatamcode:
This PR removes the Stateful Ruler StatefulSets, for both metric-federation and observatorium.

PR #208, opened 2022-04-28
Revert "Remove stateful rulers (#184)"

moadz:
This reverts commit 04b0c0bc

The plan now is to keep running stateless-ruler in parallel, continue to gather data and understanding of performance over time. Before completing the migration completely, we should have a set of SOP's/Alerts that tell us when things are going wrong. 

cc: @saswatamcode for review


saswatamcode (approved):
Thanks!

Just need to add `--rule-file` args back!

saswatamcode (approved):

PR #209, opened 2022-05-03
Add obsctl to CI for setting RHOBS rules

saswatamcode:
This PR adds obsctl to GitHub Actions for setting Rules in RHOBS as rhobs tenant in staging. 

This would serve as an example for users who want to maintain their Rule files in a repo and run obsctl in their respective CIs.

TODO: Add secrets to repo and access them as env variables.


bwplotka:
Nice! Can we move this forward?

PR #210, opened 2022-05-05
Remove rule files from stateless Rulers

saswatamcode:


services/observatorium-metrics.libsonnet
```diff
@@ -281,12 +272,6 @@ local oauthProxy = import './sidecars/oauth-proxy.libsonnet';
         name: metricFederationStatelessRuler,
         key: statelessRulerKey,
       },
-      rulesConfig: [
```
moadz, line 284: If we have the same rules in fed/non-fed and both stateless rulers replicas, we are 4x applying rules correct?

moadz (approved):
LGTM

services/observatorium-metrics.libsonnet
saswatamcode: Yes, which is what I'm removing. So we should now have 2x for stateful!

PR #211, opened 2022-05-05
Bumping readiness probe for stateless rulers to account for WAL replay

moadz:
Moving from 180s with delay of 60s. This will delay all rollouts as readiness probe is less-frequent after a longer elapsed time. We can revisit this if it mitigates ruler crashloop


saswatamcode (approved):
LGTM

PR #212, opened 2022-05-05
Bumping liveness probe for stateless rulers to account for WAL replay…

moadz:
… to 120s and 10 failure threshold


saswatamcode (approved):
LGTM

PR #218, opened 2022-05-17
IR-254: Collecting registry and image stream usage

ricardomaraschini:
As an effort to evaluate how often customers are using Image Streams and
the internal registry this commit introduces two new metrics to be
collected:

- The amount of Image Stream Tags in the cluster
- The amount of pushes and pulls executed against the registry

These new metrics are being collected through:

https://github.com/openshift/cluster-monitoring-operator/pull/1668

Signed-off-by: Ricardo Maraschini <rmarasch@redhat.com>


bwplotka (approved):
Looks good ot me - I have those are individual metrics only.

PR #232, opened 2022-06-02
add elasticsearch subscription

frzifus:
~i would like to create an elasticsearch default instance that backs up the data for different tenants. Where exactly do I do this best in rhobs? In observatorium there is something like [examples/local/main.jsonnet](https://github.com/observatorium/observatorium/blob/main/configuration/examples/local/main.jsonnet).~

Whats the prefered way, should i add my elasticsearch deployment in `services/observatorium-traces.libsonnet` or create `elasticsearch.libsonet` in [observatorium](https://github.com/observatorium/observatorium) and import it?

### DIFF
--- a/resources/services/observatorium-traces-template.yaml
+++ b/resources/services/observatorium-traces-template.yaml
@@ -3,30 +3,6 @@ kind: Template
 metadata:
   name: observatorium-traces
 objects:
-- apiVersion: operators.coreos.com/v1alpha1
-  kind: Subscription
-  metadata:
-    name: rhobs-opentelemetry
-    namespace: ${OPENTELEMETRY_OPERATOR_NAMESPACE}
-  spec:
-    channel: stable
-    installPlanApproval: Automatic
-    name: opentelemetry-product
-    source: ${OPENTELEMETRY_OPERATOR_SOURCE}
-    sourceNamespace: openshift-marketplace
-    startingCSV: opentelemetry-operator.v${OPENTELEMETRY_OPERATOR_VERSION}
-- apiVersion: operators.coreos.com/v1alpha1
-  kind: Subscription
-  metadata:
-    name: rhobs-jaeger
-    namespace: ${JAEGER_OPERATOR_NAMESPACE}
-  spec:
-    channel: stable
-    installPlanApproval: Automatic
-    name: jaeger-product
-    source: ${JAEGER_OPERATOR_SOURCE}
-    sourceNamespace: openshift-marketplace
-    startingCSV: jaeger-operator.v${JAEGER_OPERATOR_VERSION}
 - apiVersion: jaegertracing.io/v1
   kind: Jaeger
   metadata:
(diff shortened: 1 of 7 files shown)


services/observatorium-traces.libsonnet
```diff
@@ -52,5 +52,22 @@ local tracing = (import 'github.com/observatorium/observatorium/configuration/co
...
+
+    elasticsearch:: {
+      apiVersion: 'operators.coreos.com/v1alpha1',
+      kind: 'Subscription',
+      metadata: {
+        name: 'rhobs-elastic',
+        namespace: '${ELASTIC_OPERATOR_NAMESPACE}',
+      },
+      spec: {
+        channel: '5.0',  // or stable?
```
frzifus, line 64: Got the channel and other details from an openshift instance.

resources/services/observatorium-traces-template.yaml
```diff
@@ -115,3 +127,9 @@ parameters:
   value: openshift-operators
 - name: JAEGER_OPERATOR_SOURCE
   value: redhat-operators
+- name: ELASTICSEARCH_OPERATOR_VERSION
+  value: 5.0.12-1
```
pavolloffay, line 131: Is this the latest version as of now?

services/observatorium-traces.libsonnet
pavolloffay: IIRC we should use the latest stable channel 

Screenshot of Operator Installation · Red Hat OpenShift Dedicated (1)
Screenshot of OperatorHub · Red Hat OpenShift Dedicated

esnible:
I'd like to see the operators deployed in a different .jsonnet file because I'd like the OpenShift Template for creating an instance to be separate from installing the operator that does the work.

- Having them in the same file might make sense for upgrading the operator, but if the operator isn't there the CR will be of an unknown type and there will be a partial failure applying the template
- I'm testing and maybe I don't have permission to install operators (that are in fact already installed) and there might be a partial failure.

From an `oc process` workflow partial failures are confusing to script.  I don't know how Qontract handles these things.

frzifus:
> Also please test by hand that the subscription can be created in the cluster.

tested
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: rhobs-elasticsearch
  namespace: openshift-operators
spec:
  channel: stable
  installPlanApproval: Automatic
  name: elasticsearch-operator
  source: redhat-operators
  sourceNamespace: openshift-marketplace
  startingCSV: elasticsearch-operator.5.4.1-24
```

> I'd like to see the operators deployed in a different .jsonnet file because I'd like the OpenShift Template for creating an instance to be separate from installing the operator that does the work.

ok, right now the operator is not part of the pr. But that does mean `"or create elasticsearch.libsonet in observatorium and import it?"` is the way to do it?

pavolloffay:
The installation of the operator should be here in this repository and probably as well the ES CR. We don't want to make observatorium openshift specific

frzifus:
tested it on the testing cluster and it worked fine.
```bash
get pods -n openshift-operators
NAME                                                             READY   STATUS    RESTARTS   AGE
elasticsearch-operator-6967676676-p54cp                          2/2     Running   0          32s
```

resources/services/observatorium-traces-template.yaml
```diff
@@ -27,6 +27,18 @@ objects:
     source: ${JAEGER_OPERATOR_SOURCE}
     sourceNamespace: openshift-marketplace
     startingCSV: jaeger-operator.v${JAEGER_OPERATOR_VERSION}
+- apiVersion: operators.coreos.com/v1alpha1
```
pavolloffay, line 30: The jsonnet was split but also need to render the templates into a separate OCP template file.

frzifus: ok, now its
```bash
oc-4.10 apply -f resources/services/observatorium-traces-subscriptions-template.yaml
oc-4.10 apply -f resources/services/observatorium-traces-template.yaml
```

pavolloffay (approved):

periklis (approved):

PR #237, opened 2022-06-10
Remove metric-fed ruler from Querier

saswatamcode:
This PR removes metric-fed ruler from observatorium-mst Querier to stop the spam of "no such host" logs.

### DIFF
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -1047,8 +1047,6 @@ objects:
           - --store=dnssrv+_grpc._tcp.observatorium-thanos-receive-default.${NAMESPACE}.svc.cluster.local
           - --rule=dnssrv+_grpc._tcp.observatorium-thanos-rule.${NAMESPACE}.svc.cluster.local
           - --rule=dnssrv+_grpc._tcp.observatorium-thanos-stateless-rule.${NAMESPACE}.svc.cluster.local
-          - --rule=dnssrv+_grpc._tcp.observatorium-thanos-metric-federation-rule.${NAMESPACE}.svc.cluster.local
-          - --rule=dnssrv+_grpc._tcp.observatorium-thanos-metric-fed-stateless-rule.${NAMESPACE}.svc.cluster.local
           - --web.prefix-header=X-Forwarded-Prefix
           - --query.timeout=15m
           - --query.lookback-delta=15m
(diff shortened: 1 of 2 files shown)


matej-g:
@saswatamcode Is this correct? Do we want to federate into MST? 

To me it looks like we accidentally added the federation ruler to the `--rule` enpodint list in this PR: https://github.com/rhobs/configuration/pull/208/files#diff-08c043e75a0ae296db7120a1952f4272a89f4897e3af5c8433af63a6bf814423


When in fact those are provide via SD file, see my previous PR: https://github.com/rhobs/configuration/pull/103 and corresponding AppInterface config: https://gitlab.cee.redhat.com/service/app-interface/blob/master/data/services/rhobs/observatorium/observatorium-metrics/cicd/saas.yaml#L207. With current setup, we're duplicating the ruler entries.

saswatamcode:
Yes! Let me update!

moadz:
Just noting down my convo with Saswata, but I believe we should either get rid of file-based SD or get rid of arg-based SD and not keep both. My understanding is that we're going to remove file-sd via app-interface, and only define SD targets via this repo + jsonnet.

saswatamcode:
+1, to avoid further confusion it would be better to define it here. But observatorium-mst isn't fully separate within this repo's jsonnet, so it's a bit trickier to do it here.

matej-g:
If you'll take a look at my original PR, you should understand why we cannot do it without file SD - the template is shared between Observatorium and MST namespace, meaning we cannot choose different arguments for different environments (hence we'd be getting spammed in MST with the error about targets which do not exist in MST namespace). Adding extra targets via file SD was a workaround.

saswatamcode:
Totally agree! We share templates and the actual difference lies in app-interface params. But this is why it's confusing when such config is updated. 😅 

I wonder if there's a better way here? 🤔

matej-g:
@saswatamcode I will 100% support any better solution, believe me, this was an after the fact fix, when I realized we otherwise cannot prevent the failed lookup spam in logs (and the change was already in the configuration repo).

matej-g:
So what shall we do here :disappointed:

saswatamcode:
I'm also not sure here! Let me look into this more! 🙂

matej-g:
Ping, what shall we do here?

saswatamcode:
I haven't really found any better fix for this other than the current implementation. 😞

matej-g:
Then shall we remove the duplication introduced previously as I mentioned in https://github.com/rhobs/configuration/pull/237#issuecomment-1153626507?

saswatamcode:
Yes let me update this PR!

matej-g:
Ah, sorry, I got confused, now I see it's updated, all good :bow:. Let me review.

matej-g (approved):
OK, looks good :raised_hands:

PR #247, opened 2022-06-21
Moved RBAC configuration to mimic (Go code).

bwplotka:
Enough it's enough: We have inconsistency across all tenancy configurations due to a lack of proper templating. When I tried to use jsonnet to abstract things, it takes forever and looks quite awful:

image

So I decided to give the good old [mimic](https://github.com/bwplotka/mimic) project a try (it's ultra simple project, just generation nothing else).

Only for RBAC for now, not planning anything for Kubernetes YAMLs or full jsonnet replacement yet (and probably even for mid-term).

I researched Pulumi too, but it's really [too mixed up with applying](https://cloudolife.com/2021/04/10/Infrastructure-as-Code-IaC/Pulumi/Pulumi-Generate-Kubernetes-YAML-with-Golang-Go/), which can cause a lot of troubles, e.g. you don't want to log in to some k8s only to generate your Yamls. IMO we have to keep those things separate (templating vs applying)

NOTE: We can abstract this, even more, it's just an example.

Our slack discussion: https://coreos.slack.com/archives/G79AW9Q7R/p1655810892943989

### Pros and Cons

Pros:
* Mimic is just a tiny Go runnable script. If we choose to move to anything else (kustomize, Cue, Assembly) - we can change it overnight.
* Automatically builds before jsonnet
* Type safe
* Directly imports CONFIG type from https://github.com/observatorium/api RBAC which reduces the risk of producing bugs due to some typos that will be only found out when first deploy.
* All team members are familiar with Go
* It is actually helpful if we would create an operator someday (we would most likely use Golang Operator SDK).
* Testable

Cons:
* Controversial "IT FEELS WEIRD TO CONFIGURE STUFF IN GOLANG"
* There are certain [best practices to learn with configuration in Go](https://github.com/bwplotka/mimic/blob/main/docs/best_practice.md) - based on our learnings from prev job. TL;DR:
  * panics only (no `if err != nil`)
  * no side effects (reading from input, files)
  * no complex abstractions 
* Actually more chars than typing all YAML by hand (some boilerplate)


matej-g (commented):
I'm happy to give this change a go :+1: For me definitely improvement over current state.

configuration/observatorium/tenants.go
```diff
@@ -0,0 +1,228 @@
...
+)
+
+func GenerateTenants(g *mimic.Generator) {
+	obsRBAC := ObservatoriumRBAC{
+		mappedRoleNames: map[roleMapKey]string{},
+	}
+
+	// CNV-QE
+	attachBinding(&obsRBAC, bindingOpts{
+		name:    "observatorium-cnv-qe", // NOTE(bwplotka): Name is strongly correlated to service account name. Any change, require changes on tenant side.
```
douglascamata, line 47: Maybe this comment deserves to be in the definition of the `bindingOpts` type?

configuration/observatorium/tenants.go
```diff
@@ -0,0 +1,228 @@
...
+
+type env string
+
+const (
+	testingEnv    env = "testing"
+	stagingEnv    env = "staging"
+	productionEnv env = "production"
+)
+
+func GenerateTenants(g *mimic.Generator) {
```
douglascamata, line 40: As this is exported, can it have a comment to describe the function?

configuration/observatorium/tenants.go
```diff
@@ -0,0 +1,228 @@
...
+	attachBinding(&obsRBAC, bindingOpts{
+		name:    "observatorium-odfms",
+		tenant:  odfmsTenant,
+		signals: []signal{metricsSignal},
+		perms:   []rbac.Permission{rbac.Write, rbac.Read},
+		envs:    []env{stagingEnv},
+	})
+
+	// Use JSON because we want to have jsonnet using that in configmaps/secrets.
+	g.Add("rbac.json", encoding.JSON(obsRBAC))
```
douglascamata, line 158: `g` is defined more than 100 lines above in this function (which has a tendency to get bigger as we onboard new tenants) and it's difficult to remember where it comes from or what it is without having to scroll all the way up or relying on code editor features. Can we rename it to `generator`?

configuration/observatorium/tenants.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+type roleMapKey struct {
+	tenant tenantID
+	signal signal
+	perm   rbac.Permission
+}
+
+type ObservatoriumRBAC struct {
```
douglascamata, line 167: Missing comment for exported type.

configuration/observatorium/tenants.go
bwplotka: 👍🏽

configuration/observatorium/tenants.go
bwplotka: Generally I would agree, but not here. We will use generator a lot with this framework - it's a core functionality that every function will follow (it might pass `g`). I would love to keep it simple then. Maybe `gen` would suffice ? (:

douglascamata: `gen` suffices 👍

douglascamata (approved):
LGTM :shipit:

PhilipGough (approved):

PR #256, opened 2022-06-28
Adds env and namespace aware API Secret template

PhilipGough:


PhilipGough:
@bwplotka yes, there appears to be a method now in app-interface to pipe secrets from vault into the template. I just want to try this out with the test cluster.

Hopefully this pattern for splitting resources by cluster and namespace is ok?

PR #258, opened 2022-07-04
MON-2470: Define tenants.yaml config 

bill3tt:
This is the first step in migrating our tenants.yaml away from Vault & AppInterface into this repo.

When we are fully defining and shippiung the configuration from this repo, we will:
* Not be required to do the full AppSRE / PGP dance to update our tenants.yaml file.
* Be able to fully self-service changes to this file via AppInterface saas files.

There may be a couple more iterations on this file as we figure out the correct upstream parameters in AppInterface, but this is the starting point :muscle: 


resources/services/observatorium-tenants-mst-template.yaml
```diff
@@ -0,0 +1,109 @@
...
+kind: Template
+metadata:
+  name: observatorium
+objects:
+- apiVersion: v1
+  data:
+    client-id: ${CLIENT_ID}
+    client-secret: ${CLIENT_SECRET}
+    issuer-url: https://sso.redhat.com/auth/realms/redhat-external
+    tenants.yaml:
```
maorfr, line 11:
```suggestion
    tenants.yaml: |
      ${TENANTS}
```

resources/services/observatorium-tenants-template.yaml
```diff
@@ -0,0 +1,21 @@
...
+  name: ${SECRET_NAME}
+objects:
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: ${SECRET_NAME}
+  stringData:
+    client-id: ${CLIENT_ID}
+    client-secret: ${CLIENT_SECRET}
+    issuer-url: https://sso.redhat.com/auth/realms/redhat-external
```
maorfr, line 13: you can use the `SSO_BASE_URL` parameter instead.

PR #260, opened 2022-07-05
Move reconcile annotation onto the secret instead of the template

bill3tt:


douglascamata (approved):
:shipit:

PR #267, opened 2022-07-12
Add new RHODS staging accounts

douglascamata:
I confirmed the new accounts created by RHODS for staging are working fine, so I am adding them.

Will remove the old one when I can confirm it's not being used anymore.

### DIFF
--- a/configuration/observatorium/rbac.go
+++ b/configuration/observatorium/rbac.go
@@ -77,16 +77,6 @@ func GenerateRBAC(gen *mimic.Generator) {
 		perms:   []rbac.Permission{rbac.Read},
 		envs:    []env{stagingEnv},
 	})
-	// TODO(douglascamata): Old RHODS account to be removed when the 2 above are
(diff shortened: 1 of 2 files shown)


bill3tt (changes requested):
Honestly I would just remove the old one now. This is staging and we will almost certainly forget to remove that once we've gone through the process of checking with the tenant.

It's like a bandaid. Better to rip it off, than slowly pull it off bit by bit. Less pain for everyone!

(happy to be convinced otherwise tho if you feel strongly)

douglascamata:
If I put myself on the tenant's shoes, I would not like to have my only account that currently works removed from the system. That's not a good experience. 

I am taking ownership of cleaning this up later and actually rather sooner than later. :)

douglascamata:
@bill3tt I confirmed they aren't using that account yet, so I removed it.

bill3tt (approved):

PR #268, opened 2022-07-13
Removing staging postfix from rhods account

douglascamata:
This postfix is added by the templater.

I also added a verification in the RBAC file generation to prevent this from ever happening again.


jessicalins (approved):
LGTM :rocket:

PR #269, opened 2022-07-13
Allow Thanos Compact to ignore out of order labels

douglascamata:
This should help our pod start in the MST staging instance. 


bwplotka (approved):
):

douglascamata:
> Can we remove broken block instead?

I don't know how to identify and remove it. We have to find a volunteer.

douglascamata:
Going to merge and move this to staging, followed up by a removing of the flag. If the blocks causes issues we will get rid of it from the object storage.

PR #270, opened 2022-07-13
Add rhobs logs only tenants

periklis:
- Tenant DPTP is migrated from observatorium.api to observatorium-mst.api
- Tenant OCM is new for the PoC with AppSRE OCM team.

PR #277, opened 2022-07-15
[rhobs-logs] Remove obsolete ams mappings for dptp and ocm tenants

periklis:
The logs-only tenants DPTP and OCM were migrated into static rbac by #270 and this PR is a follow-up to cleanup AMS mappings.


matej-g (approved):
Good to go once https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/43463/diffs is in! Thanks for the clean-up :broom:

PR #284, opened 2022-07-19
[rhobs-logs] Disable tracing to temporarily stop panic errors

periklis:
We are currently observing panic errors on Loki components for all sampled request by Jaeger. The root cause for this issue is when using the UserID and OrgID in exemplars. See:
- grafana/loki/issues/6667
- weaveworks/common/pull/246

image

The following PR disables trace sampling for all Loki components.


periklis:
> Is this the minimum set of things we can disable to make this work? Wouldn't it be enough to just disable tracing through its boolean flag and leave all the configuration there?

Sure we can disable the flag only, but I though why keeping the sidecars idling if they don't sample anything.

douglascamata (approved):
🚢

PR #287, opened 2022-07-20
Add post deploy job for logs

periklis:
Adds a new template for a post-depoy job for the RHOBS logs endpoints.


bill3tt:
@periklis am I understanding this correctly that we hand-craft the templates rather than generate them via jsonnet?

periklis:
> @periklis am I understanding this correctly that we hand-craft the templates rather than generate them via jsonnet?

To the best of my knowledge yes it  looks we handcraft them.

bill3tt:
Let's just quickly get @matej-g to sign-off as the original author of the post-deploy jobs.

tests/integration_tests/post-deploy-logs-job-template.yaml
```diff
@@ -0,0 +1,82 @@
...
+        initContainers:
+        - command:
+          - /bin/sh
+          - -c
+          - |
+            cat > /var/logs-file/logs.yaml << EOF
+            spec:
+              logs: [ [ "$(date '+%s%N')", "log line"] ]
+            EOF
+          image: docker.io/debian
```
matej-g, line 27: Just FYI I think AppSRE has policy against using DockerHub (not familiar with `docker.io`), I assume they only allow Red Hat / Quay.io repositories, but you can double check.

matej-g (commented):
Looking good :+1: @bill3tt since there was only one test file, I did not see the need to template this (yet). Two seems still fine but if we'll find we want more tests, it might sense to think of templating.

One thing I'm thinking if we should also be reusing `runtest.sh`, since I'm planning to add few more (non-`up`) tests, e.g. to test rules API. But we can cross that bridge (split to separate test scripts) once we come to that.

bill3tt (approved):
Ship it :)

matej-g (approved):
Awesome job! :100:

PR #288, opened 2022-07-20
Parca updates

PhilipGough:
1. Bumps the default version
2. Aligns CPU limits with what we see during load
3. Makes data retention time configurable


moadz (approved):
LGTM thanks for this

PR #289, opened 2022-07-20
Fix breaking change between Parca version

PhilipGough:
There is a breaking change between 0.11.x and 0.12.0
TSDB has been removed in favour of column storage
so corresponding flag has been removed.


moadz (approved):
LGTM

PR #293, opened 2022-08-01
[rhobs] Rename ocm tenant as appsre

periklis:
OCM Team decided to rename to tenant `appsre`. This tenant was and is still unused so renaming has no impact on any system.


bwplotka (approved):
💪🏽

PR #295, opened 2022-08-16
Bump rules-objstore replicas to 2

saswatamcode:
This PR increases the number of rules-objstore replicas to 2, for better resiliency.


jessicalins (approved):
:rocket:

PR #298, opened 2022-08-23
Add Rules and Alertmanager SLOs

jessicalins:
This PR adds availability SLOs for Rules (/rules and /rules/raw) and Observatorium Alertmanager, for Telemeter and MST instances. Ticket: https://issues.redhat.com/browse/MON-2474 

~~The panels for /rules/raw have no data yet, because it depends on the handler label added in https://github.com/observatorium/api/pull/361 to be deployed.~~
I think also we'll have better measurements once we have more data (provided by synthetic data generation)

I'll add a separate PR for alerts to make it easier to review.

### DIFF
--- a/observability/dashboards/slo.libsonnet
+++ b/observability/dashboards/slo.libsonnet
@@ -510,8 +510,8 @@ function(instanceName, environment, dashboardName) {
     availabilityRow(
       '95% of alerts are successfully delivered to upstream targets',
       0.95,
-      'sum(alertmanager_notifications_failed_total{service="observatorium-alertmanager", namespace="%s"})' % instance.upNamespace,
-      'sum(alertmanager_notifications_total{service="observatorium-alertmanager", namespace="%s"})' % instance.upNamespace,
+      'sum(rate(alertmanager_notifications_failed_total{service="observatorium-alertmanager", namespace="%s"}[28d]))' % instance.upNamespace,
+      'sum(rate(alertmanager_notifications_total{service="observatorium-alertmanager", namespace="%s"}[28d]))' % instance.upNamespace,
       14
     ),
   local apiLogsPanels =
--- a/observability/dashboards/slo.libsonnet
+++ b/observability/dashboards/slo.libsonnet
@@ -479,8 +479,8 @@ function(instanceName, environment, dashboardName) {
     availabilityRow(
       '95% of rules are successfully synced to Thanos Ruler',
       0.95,
-      'sum(rate(client_api_requests_total{client="oauth",namespace="%s",code=~"5.+"}[28d]))' % instance.upNamespace,
-      'sum(rate(client_api_requests_total{client="oauth",namespace="%s",code!~"4.+"}[28d]))' % instance.upNamespace,
+      'sum(rate(client_api_requests_total{client="oauth",container="thanos-rule-syncer",namespace="%s",code=~"5.+"}[28d]))' % instance.upNamespace,
+      'sum(rate(client_api_requests_total{client="oauth",container="thanos-rule-syncer",namespace="%s",code!~"4.+"}[28d]))' % instance.upNamespace,
       10
     ) +
     titleRow('API > Rules Read (/rules) > Availability') +
--- a/observability/dashboards/slo.libsonnet
+++ b/observability/dashboards/slo.libsonnet
@@ -467,6 +467,55 @@ function(instanceName, environment, dashboardName) {
       'sum(rate(up_custom_query_duration_seconds_count{namespace="%s",query="query-path-sli-100M-samples"}[28d]))' % instance.upNamespace,
       8
     ),
+  local apiRulesPanels =
+    titleRow('API > Rules Write (/rules/raw) > Availability') +
+    availabilityRow(
+      '95% of valid write requests return successfully',
+      0.95,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules-raw", code=~"5.+", method=~"PUT"}[28d]))' % instance.apiJob,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules-raw", code!~"4.+", method=~"PUT"}[28d]))' % instance.apiJob,
+      9
+    ) +
+    titleRow('API > Rules Write > Availability') +
+    availabilityRow(
+      '95% of rules are successfully synced to Thanos Ruler',
+      0.95,
+      'sum(thanos_rule_loaded_rules{container="thanos-rule",namespace="%s"}) < 0' % instance.upNamespace,
+      'sum(prometheus_rule_group_rules{job="observatorium-thanos-rule",namespace="%s"})' % instance.upNamespace,
+      10
+    ) +
+    titleRow('API > Rules Read (/rules) > Availability') +
+    availabilityRow(
+      '90% of valid requests return successfully',
+      0.9,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules", code=~"5.+"}[28d]))' % instance.apiJob,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules", code!~"4.+"}[28d]))' % instance.apiJob,
+      11
+    ) +
+    titleRow('API > Rules Read (/rules/raw) > Availability') +
+    availabilityRow(
+      '90% of valid requests return successfully',
+      0.9,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules-raw", code=~"5.+"}[28d]))' % instance.apiJob,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules-raw", code!~"4.+"}[28d]))' % instance.apiJob,
+      12
+    ),
+  local apiAlertingPanels =
+    titleRow('API > Alerting > Availability') +
+    availabilityRow(
+      '95% of alerts are successfully delivered to Alertmanager',
+      0.95,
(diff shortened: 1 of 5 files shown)


observability/dashboards/slo.libsonnet
```diff
@@ -466,6 +466,53 @@ function(instanceName, environment, dashboardName) {
...
+      '95% of alerts are successfully delivered to Alertmanager',
+      0.95,
+      'sum(rate(thanos_alert_sender_alerts_dropped_total{container="thanos-rule",namespace="%s"}[28d]))' % instance.upNamespace,
+      'sum(rate(thanos_alert_sender_alerts_sent_total{container="thanos-rule",namespace="%s"}[28d]))' % instance.upNamespace,
+      13
+    ) +
+    availabilityRow(
+      '95% of alerts are successfully delivered to upstream targets',
+      0.95,
+      'sum(alertmanager_notifications_failed_total{service="observatorium-alertmanager", namespace="%s"})' % instance.upNamespace,
```
jessicalins, line 513: I'm not sure if here we want `sum(rate(..))` instead. When using rate I noticed the availability was around ~20%, not sure if that because we don't have a lot of data here..

saswatamcode: So I think we do need `sum(rate(..))` here, as only `sum` would give us the number of failed alerts at a particular instant, but not over a period of time which we need for SLO. We also need to adjust for breaks in the counter (via pod restarts) which `rate` handles. Except for that, this query seems to work! 🙂

observability/dashboards/slo.libsonnet
```diff
@@ -466,6 +466,53 @@ function(instanceName, environment, dashboardName) {
...
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules-raw", code=~"5.+", method=~"PUT"}[28d]))' % instance.apiJob,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules-raw", code!~"4.+", method=~"PUT"}[28d]))' % instance.apiJob,
+      9
+    ) +
+    titleRow('API > Rules Write > Availability') +
+    availabilityRow(
+      '95% of rules are successfully synced to Thanos Ruler',
+      0.95,
+      'sum(thanos_rule_loaded_rules{container="thanos-rule",namespace="%s"}) < 0' % instance.upNamespace,
+      'sum(prometheus_rule_group_rules{job="observatorium-thanos-rule",namespace="%s"})' % instance.upNamespace,
```
saswatamcode, line 483: This query seems to have a couple of problems,

- The error case query i.e, `sum(thanos_rule_loaded_rules{container="thanos-rule",namespace="%s"}) < 0` does not return data and needs a `bool` operand at the end of the query like ` }} < bool 0`. But even then it uses a gauge metric that returns the number of rules that Thanos Ruler has loaded. This gauge is never negative, so the query would always return false or 0. So availability would always end up at 100 I think.
- We need `sum(rate(..))` here too, as we are measuring SLO over 28d rolling window

I think the `95% of rules synced to Thanos Ruler` SLO might be a bit confusing, to begin with. We do not really have a concept of "failed sync" for any Rule or RuleGroup, just for Rule files, and that information is already represented by `95% of valid write requests return successfully` SLO. There is no correct error case query for this maybe.

WDYT? Should we change the SLO to `95% rule files are successfully synced to Thanos Ruler` and use/add a metric to thanos-rule-syncer for failed/succesful syncs and use that for this SLO instead? 🙂

saswatamcode (commented):
Thanks for awesome work!
LGTM aside from the rules synced and upstream alerts queries! 🌟

observability/dashboards/slo.libsonnet
jessicalins: > I think the 95% of rules synced to Thanos Ruler SLO might be a bit confusing, to begin with. We do not really have a concept of "failed sync" for any Rule or RuleGroup, just for Rule files, and that information is already represented by 95% of valid write requests return successfully SLO. There is no correct error case query for this maybe.

that is true, it will indeed be more accurate if we can measure this directly in thanos-rule-syncer! Let me add a metric there and update here once it is merged. I'm not sure though if we would need an extra metric in rules-objstore?

saswatamcode: rules-objstore already has metrics to help with the "total" part of the SLO query with metric like "rules_objstore_rules_configured". We need a new metric for "error" part of SLO query, something like `rules_objstore_failed_rules` (which can be incremented when objstore request fails).

I think having both thanos-rule-syncer and rules-objstore would be preferrable, as these are both critical path services for Rules API! So any errors on either affect SLO. 🙂

jessicalins: we discussed this a bit ^ and decided to clarify this SLO by renaming it the row to API > Rules **Sync** > Availability, since this is different than a write. We won't need then an extra metric in rules-objstore since the write is already covered by `95% of valid write requests return successfully` SLO as @saswatamcode mentioned.

observability/dashboards/slo.libsonnet
```diff
@@ -466,22 +466,69 @@ function(instanceName, environment, dashboardName) {
...
+      0.95,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules-raw",code=~"5.+",method=~"PUT"}[28d]))' % instance.apiJob,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules-raw",code!~"4.+",method=~"PUT"}[28d]))' % instance.apiJob,
+      9
+    ) +
+    titleRow('API > Rules Sync > Availability') +
+    availabilityRow(
+      '95% of rules are successfully synced to Thanos Ruler',
+      0.95,
+      'sum(rate(client_api_requests_total{client="oauth",namespace="%s",code=~"5.+"}[28d]))' % instance.upNamespace,
```
jessicalins, line 482: @saswatamcode I was researching a bit and I think we can actually use the `client_api_requests_total` metric that is already [present](https://github.com/observatorium/thanos-rule-syncer/blob/main/instrument.go#L19) in `thanos-rule-syncer`.

At first I thought about adding a new metric there at the end of the 'sync' process, at the end of the reload [function](https://github.com/observatorium/thanos-rule-syncer/blob/main/main.go#L179).

But then looking a bit more, the same `oauth` client is used in all requests performed by `thanos-rule-syncer` (as far as I can understand), since is used as `Transport` [here](https://github.com/observatorium/thanos-rule-syncer/blob/main/main.go#L94).

So in all the requests that `thanos-rule-syncer` does ((`getRules` either using the rules backend URL or the observatorium API URL) + calling the reload endpoint) I think we are using the same client, with the same `Transport`, and this is already instrumented. So I think if one of those requests throws a 'bad' status code, means that then `thanos-rule-syncer` could not sync the rules correctly.

I've updated the SLO query to use this metric, now it is similar to the other availability SLO queries - what do you think?

saswatamcode: I think that could work! The only thing would be adding a `container="thanos-rule-syncer"` label to the metrics in query, as other services expose `client_api_requests_total` with `client="oauth"` as we see [here](https://prometheus.telemeter-prod-01.devshift.net/graph?g0.expr=client_api_requests_total%7Bclient%3D%22oauth%22%7D&g0.tab=1&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1h). 🙂

jessicalins: good catch! added it :)

saswatamcode (approved):

jessicalins:
@saswatamcode just noticed that we dont seem to be using the `--oidc.issuer-url` flag, which then it wouldn't use the instrumented `Transport` in https://github.com/observatorium/thanos-rule-syncer/blob/main/main.go#L94 
I'll leave this PR on hold now and add a change to `thanos-rule-syncer` so that we also instrument the 'default' `Transport` https://github.com/observatorium/thanos-rule-syncer/blob/main/main.go#L73 
Once this is merged I'll update here

jessicalins:
@saswatamcode this is ready for another look - I've updated the sync SLO query and also fixed the namespace for some of the queries (for telemeter the namespace varies and not necessarily `upNamespace` is always used)

saswatamcode (approved):
Thanks, this looks good to me! Let's deploy to stage! 🚀

observability/dashboards/slo.libsonnet
```diff
@@ -466,22 +469,69 @@ function(instanceName, environment, dashboardName) {
...
+      0.95,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules-raw",code=~"5.+",method=~"PUT"}[28d]))' % instance.apiJob,
+      'sum(rate(http_requests_total{job="%s",group="metricsv1",handler=~"rules-raw",code!~"4.+",method=~"PUT"}[28d]))' % instance.apiJob,
+      9
+    ) +
+    titleRow('API > Rules Sync > Availability') +
+    availabilityRow(
+      '95% of rules are successfully synced to Thanos Ruler',
+      0.95,
+      'sum(rate(client_api_requests_total{client="reload",container="thanos-rule-syncer",namespace="%s",code=~"5.+"}[28d]))' % instanceNamespace(instanceName),
```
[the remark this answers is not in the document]
saswatamcode: So as this is only for sync, I think it's fine! We can isolate that the rule-syncer is failing if this SLO ever drops. Mixing fetch and sync might make this confusing.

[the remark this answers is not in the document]
jessicalins: ok, lets merge like this and if we feel the need for another SLO just for fetch we can always update in the future!

PR #300, opened 2022-08-24
Add alerts for Rules and Alerting SLOs

jessicalins:
Adding alerts for the defined SLOs at https://github.com/rhobs/configuration/pull/298 
I think selectors may change slightly as we update https://github.com/rhobs/configuration/pull/298 but I already wanted to open for review. 
I'll then add runbooks in a separate PR, to make easier for review.

I think we can then have this merged after https://github.com/rhobs/configuration/pull/298 to stage, test and just when we have all runbooks in place and alerts are ok, have it deployed to production.

### DIFF
--- a/observability/prometheusrules.jsonnet
+++ b/observability/prometheusrules.jsonnet
@@ -370,6 +370,78 @@ local renderAlerts(name, environment, mixin) = {
           }),
         ],
       },
+      {
+        name: 'rhobs-' + instance + '-api-rules-raw-write-availability.slo',
+        slos: [
+          slo.errorburn({
+            alertName: 'APIRulesRawWriteAvailabilityErrorBudgetBurning',
+            alertMessage: 'API /rules/raw endpoint is burning too much error budget to guarantee availability SLOs',
+            metric: 'http_requests_total',
+            selectors: [apiJobSelector, 'group="metricsv1"', 'handler="rules-raw"', 'code=~"^(2..|3..|5..)$"', 'method=~"PUT"'],
+            errorSelectors: ['code=~"5.+"'],
+            target: 0.95,
+          }),
+        ],
+      },
+      {
+        name: 'rhobs-' + instance + '-api-rules-sync-availability.slo',
+        slos: [
+          slo.errorburn({
+            alertName: 'APIRulesSyncAvailabilityErrorBudgetBurning',
+            alertMessage: 'API /reload endpoint is burning too much error budget to guarantee availability SLOs',
+            metric: 'client_api_requests_total',
+            selectors: ['client="oauth"', upNamespaceSelector, 'code=~"^(2..|3..|5..)$"'],
+            errorSelectors: ['code=~"5.+"'],
+            target: 0.95,
+          }),
+        ],
+      },
+      {
+        name: 'rhobs-' + instance + '-api-rules-read-availability.slo',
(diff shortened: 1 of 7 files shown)


observability/prometheusrules.jsonnet
```diff
@@ -370,6 +370,78 @@ local renderAlerts(name, environment, mixin) = {
...
+          }),
+        ],
+      },
+      {
+        name: 'rhobs-' + instance + '-api-alerting-availability.slo',
+        slos: [
+          slo.errorburn({
+            alertName: 'APIAlertmanagerAvailabilityErrorBudgetBurning',
+            alertMessage: 'API Thanos Rule failing to send alerts to Alertmanager and is burning too much error budget to guarantee availability SLOs',
+            metric: 'thanos_alert_sender_alerts_dropped_total',
```
jessicalins, line 432: I wonder if we need this alert - we seem to use already the thanos mixin and it contains the [ThanosRuleSenderIsFailingAlerts](https://github.com/thanos-io/thanos/blob/main/mixin/alerts/rule.libsonnet#L32) alert, though I think it is not written specifically for our SLOs.

observability/prometheusrules.jsonnet
```diff
@@ -370,6 +370,78 @@ local renderAlerts(name, environment, mixin) = {
...
+        ],
+      },
+      {
+        name: 'rhobs-' + instance + '-api-rules-sync-availability.slo',
+        slos: [
+          slo.errorburn({
+            alertName: 'APIRulesSyncAvailabilityErrorBudgetBurning',
+            alertMessage: 'API /reload endpoint is burning too much error budget to guarantee availability SLOs',
+            metric: 'client_api_requests_total',
+            selectors: ['client="oauth"', upNamespaceSelector, 'code=~"^(2..|3..|5..)$"'],
```
saswatamcode, line 393: Also commented on #298, but we might want a `container` label here. :)

saswatamcode (approved):
Thanks! LGTM! 🌟 

One minor nit

observability/prometheusrules.jsonnet
```diff
@@ -288,9 +288,12 @@ local renderAlerts(name, environment, mixin) = {
     },
   ],
 
-  local apiSLOs = function(instance, upNamespace, apiJob) {
+  local apiSLOs = function(instance, upNamespace, metricsNamespace, apiJob) {
     local apiJobSelector = 'job="' + apiJob + '"',
     local upNamespaceSelector = 'namespace="' + upNamespace + '"',
+    local metricsNamespaceSelector = 'namespace="' + metricsNamespace + '"',
+    local instanceNamespace(name) = if instance == 'telemeter' then metricsNamespaceSelector else upNamespaceSelector,
```
jessicalins, line 295: I feel like there may be a better way to do this - it is currently repeating similar logic as in https://github.com/rhobs/configuration/blob/main/observability/dashboards/slo.libsonnet#L35
not sure if makes sense/is worth extracting this function to a more common place, open to ideas

saswatamcode: Maybe extracting to `configuration/` might be better? 🙂

saswatamcode (approved):
Thanks for awesome work!

saswatamcode (approved):

PR #301, opened 2022-08-25
Add obsctl-reloader for Rules API SLO

saswatamcode:
The PR adds a deployment of obsctl-reloader to observatorium-template for generating synthetic data for Rules & Alerts API SLO.

It also adds a PrometheusRule object that contains 3 Rule Groups with 7 alerting/recording rules. This PrometheusRule is not used by app-sre Prometheus, but only by obsctl-reloader.


jessicalins (approved):
LGTM :+1:

PR #305, opened 2022-09-05
Add missing for field in SLO rule

saswatamcode:

### DIFF
--- a/services/observatorium.libsonnet
+++ b/services/observatorium.libsonnet
@@ -257,6 +257,7 @@ local obsctlReloader = (import 'github.com/rhobs/obsctl-reloader/jsonnet/lib/obs
             {
               alert: 'CombinedAlert',
               expr: 'vector(1)',
+              'for': '1m',
               annotations: {
                 description: 'Combined alert firing!',
                 message: 'Combined alert fired.',
--- a/resources/services/observatorium-template.yaml
+++ b/resources/services/observatorium-template.yaml
@@ -458,6 +458,7 @@ objects:
           description: Combined alert firing!
           message: Combined alert fired.
         expr: vector(1)
+        for: 1m
         labels:
           severity: page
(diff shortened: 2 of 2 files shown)


PhilipGough (approved):

services/observatorium.libsonnet
```diff
@@ -257,6 +257,7 @@ local obsctlReloader = (import 'github.com/rhobs/obsctl-reloader/jsonnet/lib/obs
             {
               alert: 'CombinedAlert',
               expr: 'vector(1)',
+              'for': '1m',
```
moadz, line 260: is the `for` supposed to be in quotes? Rest of the object does not escape the key.

moadz (approved):

services/observatorium.libsonnet
PhilipGough: This generated ok so its not an issue but I think we should introduce a format check for jsonnet. I take an action item on it.

saswatamcode: So for is a keyword in jsonnet, so if for wasn't in quotes, generation would fail with
```
RUNTIME ERROR: services/observatorium.libsonnet:260:15-18 Object comprehension can only have one field
	services/metric-federation-rule-template.jsonnet:1:13-45	thunk <obs> from <$>
	services/metric-federation-rule-template.jsonnet:10:34-37	
	<std>:1501:24-25	thunk from <function <anonymous>>
	<std>:1501:5-33	function <anonymous>
	services/metric-federation-rule-template.jsonnet:10:17-55	object <anonymous>
	services/metric-federation-rule-template.jsonnet:(6:12)-(12:4)	
	Field "objects"	
	During manifestation	
```

PhilipGough: right, and formatting is already done via `make format` :)

PR #307, opened 2022-09-12
Add Thanos Receive migration template

PhilipGough:
These are temporary resources and not to be merged. I am just creating a draft PR for visibility and history.

These resources facilitate the migration between two sets of Thanos Receive StatefulSets in order to circumvent a breaking change in the hashing algorithm. 

Related to https://issues.redhat.com/browse/RHOBS-112

### DIFF
--- a/resources/services/receiver-migration-template.yaml
+++ b/resources/services/receiver-migration-template.yaml
@@ -89,6 +89,8 @@ objects:
   kind: ConfigMap
   metadata:
     name: mirror-proxy
+    annotations:
+      qontract.recycle: "true"
 - apiVersion: apps/v1
   kind: Deployment
   metadata:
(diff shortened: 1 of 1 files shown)


bwplotka:
Nice, I can see we boycott jsonnet ;p 

Suggestion: Use simpler names for target resources e.g. `receive` or `receive-ketama` and `receive-controller-ketama`  instead of long names.

PhilipGough:
>  I can see we boycott jsonnet

So far at least :) 

With respect to the naming conventions, they are explicit but they are also temporary, after the migration takes place, the idea is that the old resources will be reconfigured correctly and the newer ones (these) stood down. The idea being that we do not inadvertently affect any dashboards, alerting pipelines/config etc!

PR #308, opened 2022-09-13
Enable configuration of remote write endpoint via params

PhilipGough:
To support writing to an intermediate proxy, we need to be able to configure the service name and the port for the remote write endpoint.

### DIFF
--- a/resources/services/observatorium-template.yaml
+++ b/resources/services/observatorium-template.yaml
@@ -29,7 +29,7 @@ objects:
         - args:
           - --metric-count=1
           - --series-count=8333
-          - --remote-url=http://observatorium-thanos-receive.${OBSERVATORIUM_METRICS_NAMESPACE}.svc.cluster.local:19291/api/v1/receive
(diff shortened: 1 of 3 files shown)


resources/services/observatorium-template.yaml
```diff
@@ -29,7 +29,7 @@ objects:
         - args:
           - --metric-count=1
           - --series-count=8333
-          - --remote-url=http://observatorium-thanos-receive.${OBSERVATORIUM_METRICS_NAMESPACE}.svc.cluster.local:19291/api/v1/receive
+          - --remote-url=http://${METRICS_WRITE_SERVICE_NAME}.${OBSERVATORIUM_METRICS_NAMESPACE}.svc.cluster.local:${METRICS_WRITE_SERVICE_PORT}/api/v1/receive
```
jessicalins, line 32: wouldn't we need to change https://github.com/rhobs/configuration/blob/main/services/observatorium-template.jsonnet as well?

saswatamcode: We [boycott](https://github.com/rhobs/configuration/pull/307#issuecomment-1243847675) jsonnet xD?

saswatamcode (approved):
LGTM mod missing jsonnet! 🌟

resources/services/observatorium-template.yaml
PhilipGough: we cant boycott jsonnet here unfortunately..working on it :)

PR #310, opened 2022-09-13
Enable configuring receiver service name via params for query

PhilipGough:
In order to support migration between two distinct StatefulSets, we want to be able to configure the Service name via param in the Thanos Query

### DIFF
--- a/services/observatorium-metrics.libsonnet
+++ b/services/observatorium-metrics.libsonnet
@@ -566,8 +566,10 @@ local oauthProxy = import './sidecars/oauth-proxy.libsonnet';
       stores: [
         'dnssrv+_grpc._tcp.%s.%s.svc.cluster.local' % [service.metadata.name, service.metadata.namespace]
         for service in
-          [thanos.stores.shards[shard].service for shard in std.objectFields(thanos.stores.shards)] +
-          [thanos.receivers.hashrings[hashring].service for hashring in std.objectFields(thanos.receivers.hashrings)]
+          [thanos.stores.shards[shard].service for shard in std.objectFields(thanos.stores.shards)]
+      ] + [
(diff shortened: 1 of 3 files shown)


services/observatorium-metrics.libsonnet
```diff
@@ -566,8 +566,10 @@ local oauthProxy = import './sidecars/oauth-proxy.libsonnet';
       stores: [
         'dnssrv+_grpc._tcp.%s.%s.svc.cluster.local' % [service.metadata.name, service.metadata.namespace]
         for service in
-          [thanos.stores.shards[shard].service for shard in std.objectFields(thanos.stores.shards)] +
-          [thanos.receivers.hashrings[hashring].service for hashring in std.objectFields(thanos.receivers.hashrings)]
+          [thanos.stores.shards[shard].service for shard in std.objectFields(thanos.stores.shards)]
+      ] + [
+        // todo - @pgough revert after https://issues.redhat.com/browse/RHOBS-112
```
douglascamata, line 571: Maybe instead of the `TODO` we could have a ticket in JIRA with the proper link to the ticket mentioned here and assignee?

douglascamata (approved):

services/observatorium-metrics.libsonnet
PhilipGough: Yep, I'll create a follow on ticket for a full clean up, both here and in app-interface and link it to one mentioned here.

douglascamata: IMHO, you don't even need to link it here. Up to you, @PhilipGough

saswatamcode (approved):

bwplotka (approved):

PR #311, opened 2022-09-13
Add ServiceMonitor and expose metrics port to thanos-rule-syncer

jessicalins:
I noticed we don't have a metrics port exposed in `thanos-rule-syncer` - so this PR tries to add service monitor + expose metrics.

I followed the same service monitor config we do for `opa-ams` [here](https://github.com/rhobs/configuration/blob/main/services/sidecars/opa-ams.libsonnet), ~~but when running `make all` I don't see any changes in the yaml's.~~

~~I see that the `thanos-rule-syncer.libsonnet` is imported [here](https://github.com/rhobs/configuration/blob/main/services/observatorium-metrics-template-overwrites.libsonnet#L6) though, so I'm trying to understand how to integrate the serviceMonitor there, since we seem to be using just the [ruleSyncerSidecar.statefulSet](https://github.com/rhobs/configuration/blob/main/services/observatorium-metrics-template-overwrites.libsonnet#L106)~~

~~Already opening as a draft PR, to check if this is the right direction to go~~

EDIT: I paired with @moadz (thanks!) and now hopefully is the right config - we've added the servicemonitor and service to thanos rule as well - so marking this as ready for review.

### DIFF
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -2154,6 +2154,9 @@ objects:
     - name: reloader
       port: 9533
       targetPort: 9533
+    - name: thanos-rule-syncer-metrics
+      port: 8083
+      targetPort: 8083
     selector:
       app.kubernetes.io/component: rule-evaluation-engine
       app.kubernetes.io/instance: observatorium
@@ -2189,6 +2192,7 @@ objects:
         - pod
         targetLabel: instance
     - port: reloader
+    - port: thanos-rule-syncer-metrics
     namespaceSelector:
       matchNames: ${{NAMESPACES}}
     selector:
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -731,6 +731,9 @@ objects:
     - name: reloader
       port: 9533
       targetPort: 9533
+    - name: thanos-rule-syncer-metrics
+      port: 8083
+      targetPort: 8083
     selector:
       app.kubernetes.io/component: rule-evaluation-engine
       app.kubernetes.io/instance: observatorium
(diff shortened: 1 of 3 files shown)


services/observatorium-metrics-template-overwrites.libsonnet
```diff
@@ -135,6 +137,8 @@ local thanosRuleSyncer = import './sidecars/thanos-rule-syncer.libsonnet';
     },
 
     statelessRule+:: {
+      service+: ruleSyncerSidecar.service,
```
jessicalins, line 140: not sure if we want to add to stateless rule as well

saswatamcode: We can skip! :)

resources/services/observatorium-metrics-template.yaml
```diff
@@ -731,6 +731,9 @@ objects:
     - name: reloader
       port: 9533
       targetPort: 9533
+    - name: thanos-rule-syncer-metrics
+      port: 8083
```
saswatamcode, line 735: So I think this config would work, but we do not have a handler for metrics on thanos-rules-syncer. If you take a look at [main.go](https://github.com/observatorium/thanos-rule-syncer/blob/main/main.go#L66), we only run the main functionality of syncing rules and listen for os signals, but there's no server from which metrics could be scraped.

I raised https://github.com/observatorium/thanos-rule-syncer/pull/14 to add an internal server to it which exposes metrics at 8083 by default. 🙂

jessicalins: ha funny! good catch, thanks for fixing it!

services/observatorium-metrics-template-overwrites.libsonnet
jessicalins: removed:)

saswatamcode (approved):

PR #312, opened 2022-09-14
Enable hashing algorithm for receive to be set via parameter

PhilipGough:
Note, this is a breakinbg change which requires the use of Thanos >= v0.28.0


matej-g (commented):
Looks good!

PR #320, opened 2022-09-14
Enable configuration of metric fed service name and port via param

PhilipGough:


matej-g:
As @saswatamcode pointed out, we might not need this for the receiver migration, as technically they are not evaluating any rules currently.

PhilipGough:
Yeah, after investigation it seems not needed. Thanks for the input. Closing

PR #321, opened 2022-09-15
Fix memcached dashboard + Updated Telemeter allow-list

matej-g:
This PR fixes the memcached overview dashboard. The mixin we are using depend on using the `cluster` label, however we do not use `cluster` label (resulting in empty dashboards). Rather, we use `namespace` in our other dashboards and this PR replaces usage of `cluster` by `namespace` for memcached as well.

Includes also metric allow-list refresh, which seem to got out of sync.


bwplotka (approved):
LGTM! Fixed PR title to mention the allow-list update.

PR #328, opened 2022-09-26
Fix SLO alerting for metrics

saswatamcode:
This PR attempts to fix the alerts for SLOs that do not work currently as mentioned in https://issues.redhat.com/browse/RHOBS-453 and [slack thread](https://coreos.slack.com/archives/G79AW9Q7R/p1663937832667149).  This removes the `code` label completely and splits up regex label selectors. This does not touch logging SLO alerts. 

This currently isn't working due to the [SLO jsonnet library](https://github.com/metalmatze/slo-libsonnet) that we use, which converts all selectors to labels for rule configs. 

At some point, we might consider migrating off of the lib, as it is now also deprecated!

A few examples of how this is broken, 

For `code` label, take the [`TelemeterServerMetricsWriteAvailabilityErrorBudgetBurning`](https://github.com/rhobs/configuration/blob/main/resources/observability/prometheusrules/rhobs-slos-telemeter-production.prometheusrules.yaml#L20) alert which uses expression like `haproxy_server_http_responses_total:burnrate5m{route=~"telemeter-server-upload|telemeter-server-metrics-v1-receive",code=~"^(2..|3..|5..)$"}`. This would only select code labels like `code="200"`, `code="300"` or `code="500"`.
But this is a recording rule as defined [here](https://github.com/rhobs/configuration/blob/main/resources/observability/prometheusrules/rhobs-slos-telemeter-production.prometheusrules.yaml#L115) and the code label [added](https://github.com/rhobs/configuration/blob/main/resources/observability/prometheusrules/rhobs-slos-telemeter-production.prometheusrules.yaml#L119) to this rule is just the regex string. So the alert expression would not return data as shown [here](https://prometheus.telemeter-prod-01.devshift.net/graph?g0.expr=sum(haproxy_server_http_responses_total%3Aburnrate5m%7Broute%3D~%22telemeter-server-upload%7Ctelemeter-server-metrics-v1-receive%22%2Ccode%3D~%22%5E(2..%7C3..%7C5..)%24%22%7D)%20%3E%20(14.40%20*%20(1-0.95000))%0Aand%20sum(haproxy_server_http_responses_total%3Aburnrate1h%7Broute%3D~%22telemeter-server-upload%7Ctelemeter-server-metrics-v1-receive%22%2Ccode%3D~%22%5E(2..%7C3..%7C5..)%24%22%7D)%20%3E%20(14.40%20*%20(1-0.95000))&g0.tab=0&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1h) so it never fires.

We can take the same alert but look at the `route` label which is also a regex and is [used](https://github.com/rhobs/configuration/blob/main/resources/observability/prometheusrules/rhobs-slos-telemeter-production.prometheusrules.yaml#L20) as a regex label selector, which matches labels like `route="telemeter-server-upload"` or `route="telemeter-server-metrics-v1-receive"`. But for the recording rule, only the string is [added](https://github.com/rhobs/configuration/blob/main/resources/observability/prometheusrules/rhobs-slos-telemeter-production.prometheusrules.yaml#L120) as the route label. So the alert rule again cannot get data.

In some places `handler` labels also use regex, which can lead to confusing scenarios, so have split them up as well.

Feel free to point out if I'm wrong somewhere.

Also open question, on how code labels can be handled? I've removed them temporarily, but they filtered out any 400 response from all the alerts. I have some jsonnet hack https://github.com/saswatamcode/slo-libsonnet/commit/e230327cfd2bb37d5014e26c58138c3b8ce6536a which uses `ONLY_IN_BASE_` as a prefix to the selectors, to identify which selectors to include while generating rule configs. But this can be iterated on in future PR!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ $(JSONNET_VENDOR_DIR): $(JB) jsonnetfile.json jsonnetfile.lock.json
 
 .PHONY: update
 update: $(JB) jsonnetfile.json jsonnetfile.lock.json
-	@$(JB) update --jsonnetpkg-home="$(JSONNET_VENDOR_DIR)" https://github.com/metalmatze/slo-libsonnet/slo-libsonnet@master
+	@$(JB) update --jsonnetpkg-home="$(JSONNET_VENDOR_DIR)" https://github.com/saswatamcode/slo-libsonnet/slo-libsonnet@special-selector
 
 .PHONY: format
 format: $(JSONNET_SRC) $(JSONNETFMT)
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ $(JSONNET_VENDOR_DIR): $(JB) jsonnetfile.json jsonnetfile.lock.json
 
 .PHONY: update
 update: $(JB) jsonnetfile.json jsonnetfile.lock.json
-	@$(JB) update --jsonnetpkg-home="$(JSONNET_VENDOR_DIR)" https://github.com/thanos-io/kube-thanos/jsonnet/kube-thanos@main
+	@$(JB) update --jsonnetpkg-home="$(JSONNET_VENDOR_DIR)" https://github.com/metalmatze/slo-libsonnet/slo-libsonnet@master
 
 .PHONY: format
 format: $(JSONNET_SRC) $(JSONNETFMT)
--- a/jsonnetfile.json
+++ b/jsonnetfile.json
@@ -40,11 +40,11 @@
     {
       "source": {
         "git": {
-          "remote": "https://github.com/metalmatze/slo-libsonnet.git",
+          "remote": "https://github.com/saswatamcode/slo-libsonnet.git",
           "subdir": "slo-libsonnet"
         }
       },
-      "version": "master"
+      "version": "special-selector"
     },
     {
       "source": {
--- a/jsonnetfile.lock.json
+++ b/jsonnetfile.lock.json
@@ -149,8 +149,8 @@
           "subdir": "slo-libsonnet"
         }
       },
-      "version": "e238df4fac957357d78a405966c51523ef151cbc",
-      "sum": "5XrxbGgMFMHnKGeE++fiduDlDEiDunM2ZMTYCmPbWgA="
+      "version": "3db7d7d58aac44fee4006760594bce1995bed779",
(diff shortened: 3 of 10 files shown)


jessicalins (approved):
LGTM! Thanks for also fixing this to the all affected alerts with regex selectors!
Regarding the code label I think we can remove it for now and test those alerts in stage - I'm not sure if it is actually needed, but lets see :)

Makefile
```diff
@@ -22,7 +22,7 @@ $(JSONNET_VENDOR_DIR): $(JB) jsonnetfile.json jsonnetfile.lock.json
 
 .PHONY: update
 update: $(JB) jsonnetfile.json jsonnetfile.lock.json
-	@$(JB) update --jsonnetpkg-home="$(JSONNET_VENDOR_DIR)" https://github.com/thanos-io/kube-thanos/jsonnet/kube-thanos@main
+	@$(JB) update --jsonnetpkg-home="$(JSONNET_VENDOR_DIR)" https://github.com/metalmatze/slo-libsonnet/slo-libsonnet@master
```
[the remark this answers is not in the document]
saswatamcode: Ack will do!

jsonnetfile.json
```diff
@@ -40,11 +40,11 @@
     {
       "source": {
         "git": {
-          "remote": "https://github.com/metalmatze/slo-libsonnet.git",
+          "remote": "https://github.com/saswatamcode/slo-libsonnet.git",
```
matej-g, line 43: Can we just add a note or link to JIRA issue or something similar so that we understand why we need to use the fork?

matej-g (approved):
Niceee :clap:

jsonnetfile.json
saswatamcode: I don't really have a link will add a small explanation.

saswatamcode: Added a comment in the place where it's used, as I can't add comments in json. 🙂

Makefile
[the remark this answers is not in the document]
saswatamcode: Actually, let me add this in next PR and cleanup stuff a bit as well!

jessicalins (approved):
LGTM! awesome work :tada:

PR #331, opened 2022-09-26
Add Loki ruler and static rules for tenant OCM

periklis:
The following PR enables the Loki ruler component and adds three static alerting rules for the logs-only tenant OCM.

### DIFF
--- a/services/observatorium-logs-template.jsonnet
+++ b/services/observatorium-logs-template.jsonnet
@@ -22,6 +22,8 @@ local obs = import 'observatorium.libsonnet';
   ],
   parameters: [
     { name: 'NAMESPACE', value: 'observatorium-logs' },
+    { name: 'ALERTMANAGER_NAMESPACE', value: 'observatorium-mst-stage' },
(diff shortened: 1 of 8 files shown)


services/observatorium-logs-template.jsonnet
```diff
@@ -22,6 +22,8 @@ local obs = import 'observatorium.libsonnet';
   ],
   parameters: [
     { name: 'NAMESPACE', value: 'observatorium-logs' },
+    { name: 'ALERTMANAGER_NAMESPACE', value: 'observatorium-metrics' },
```
jessicalins, line 25: since `ALERTMANAGER_EXTERNAL_URL` uses `observatorium-alertmanager-mst...` shouldn't it use `observatorium-mst-stage` instead of `observatorium-metrics` as namespace? (see https://console-openshift-console.apps.app-sre-stage-0.k3s7.p1.openshiftapps.com/k8s/ns/observatorium-mst-stage/routes)

jessicalins (commented):
as far as I understand looks good - left just one question :)

jessicalins (approved):
LGTM :)

PR #336, opened 2022-09-28
Add unit tests for high sev Prometheus Rules

saswatamcode:
This PR adds unit tests for our SLO alerts.
Ideally we should setup python qontract-reconcile CI to run these tests within config itself. Will add this later on and fix CI to not remove these files. 

cc: @jessicalins


resources/observability/prometheusrules/rhobs-slos-mst-stage.prometheusrules.yaml
```diff
@@ -11,11 +11,11 @@ spec:
   groups:
   - name: rhobs-mst-api-metrics-write-availability.slo
     rules:
-    - alert: APIMetricsWriteAvailabilityErrorBudgetBurning
+    - alert: APIMetricsWriteAvailabilityErrorBudgetBurning5mand1h
```
jessicalins, line 14: there are some alerts with same name (e.g. https://github.com/rhobs/configuration/blob/047c0144d42d0caf44e38a6ed48ab1997428fd6c/resources/observability/prometheusrules/rhobs-slos-mst-stage.prometheusrules.yaml#L614) I think we'd need to change their names too?

resources/observability/prometheusrules/rhobs-slos-telemeter.prometheusrulestest.yaml
```diff
@@ -0,0 +1,236 @@
+---
```
jessicalins, line 1: I think for consistency we could use `prometheusrulestests` as name of this file, similar to the other test files

jessicalins (commented):
awesome work! :sparkle: 
I left some suggestions.
I think we could create also another issue so that we remember to set up this on CI, as you mentioned in the PR description.
I think there are still some alerts with the same name, maybe would be good to double check the other yaml files.

resources/observability/prometheusrules/rhobs-slos-telemeter-stage.prometheusrules.yaml
```diff
@@ -11,11 +11,11 @@ spec:
   groups:
   - name: rhobs-telemeter-telemeter-server-metrics-write-availability.slo
     rules:
-    - alert: TelemeterServerMetricsUploadWriteAvailabilityErrorBudgetBurning
+    - alert: TelemeterServerMetricsUploadWriteAvailabilityErrorBudgetBurning5mand1h
       annotations:
         dashboard: https://grafana.app-sre.devshift.net/d/080e53f245a15445bdf777ae0e66945d/rhobs-telemeter-telemeter-server-metrics-write-availability.slo?orgId=1&refresh=10s&var-datasource=app-sre-stage-01-prometheus&var-namespace={{$labels.namespace}}&var-job=All&var-pod=All&var-interval=5m
         message: Telemeter Server /upload is burning too much error budget to guarantee availability SLOs
-        runbook: https://github.com/rhobs/configuration/blob/main/docs/sop/observatorium.md#telemeterservermetricsuploadwriteavailabilityerrorbudgetburning
+        runbook: https://github.com/rhobs/configuration/blob/main/docs/sop/observatorium.md#telemeterservermetricsuploadwriteavailabilityerrorbudgetburning5mand1h
```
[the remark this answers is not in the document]
saswatamcode: Yup! Will make it consistent

resources/observability/prometheusrules/rhobs-slos-mst-stage.prometheusrules.yaml
saswatamcode: This is a different name, or do you mean some other alert? 🤔

resources/observability/prometheusrules/rhobs-slos-telemeter.prometheusrulestest.yaml
saswatamcode: Ack will add missing s

resources/observability/prometheusrules/rhobs-slos-mst-stage.prometheusrules.yaml
jessicalins: So I mean that we have the same alert names in different lines - for example: 

https://github.com/rhobs/configuration/blob/047c0144d42d0caf44e38a6ed48ab1997428fd6c/resources/observability/prometheusrules/rhobs-slos-mst-stage.prometheusrules.yaml#L491

https://github.com/rhobs/configuration/blob/047c0144d42d0caf44e38a6ed48ab1997428fd6c/resources/observability/prometheusrules/rhobs-slos-mst-stage.prometheusrules.yaml#L614

https://github.com/rhobs/configuration/blob/047c0144d42d0caf44e38a6ed48ab1997428fd6c/resources/observability/prometheusrules/rhobs-slos-mst-stage.prometheusrules.yaml#L737

checking now I think some alerts are in different groups, but others are in the same group (for example the ones from lines 614/737 are in the same `rhobs-mst-api-metrics-read-latency.slo` group)
I didn't check all the files but could be this is happening in other places as well - not sure if this is what we'd like to have?

saswatamcode: Yup so for this particular alert it is the same name for /query and /query_range endpoints. But desc is different.

saswatamcode:
Have added runbooks for all alerts!

jessicalins (approved):
awesome work! thanks for this :star:

PR #337, opened 2022-09-29
Add /rules/raw row to Observatorium API dashboard + refactor

jessicalins:
This PR refactors `observatorium-api.libsonnet` to reuse common parts and ideally makes it easier to add new rows/panels to the Observatorium API grafana dashboard.

I've added also a new row for `/rules/raw` - so it makes easier to debug once alerts fire for our Rules path (for example when following [this](https://github.com/rhobs/configuration/blob/main/docs/sop/observatorium.md#apirulesrawwriteavailabilityerrorbudgetburning) runbook)

~~Testing dashboard [here](https://grafana.stage.devshift.net/d/Tg-mH0rizaSJDKSADx/test2-api?orgId=1&refresh=1m&from=now-24h&to=now)~~ 
EDIT: Updated testing dashboard with error codes [here](https://grafana.stage.devshift.net/d/Tg-mH0rizaSJDKSADa/test1-api?orgId=1&refresh=1m)
 
Availability is low for /rules/raw read since we don't have traffic reading this endpoint. I think next step would be re adapt obsctl-reloader to constant 'read' this endpoint and generate constant traffic there


douglascamata:
From looking at the test dashboard, in the errors chart we're using the caption `"Errors"` for all the series returned by the query `sum(rate(http_requests_total{job="observatorium-observatorium-api",handler="rules-raw",code=~"5.."}[5m])) / sum(rate(http_requests_total{job="observatorium-thanos-query",handler="rules-raw"}[5m]))`. 

I think would be very useful to include the `code` label in the caption to help out with debugging. What do you think? 

This is a feedback that I have regarding many of our dashboards and I plan to start similar discussions/changes everywhere that it's applicable.

jessicalins:
@douglascamata that's a good idea! let me try adding this in this PR for this dashboard. In case this applies to other dashboards, maybe we could add a ticket/issue so that this is not forgotten?

douglascamata:
FYI I created an issue for the other dashboards widgets: https://github.com/rhobs/configuration/issues/348

observability/dashboards/observatorium-api.libsonnet
```diff
@@ -336,8 +336,8 @@ function(datasource, namespace) {
   local targetsAllQueryErr =
     [
       {
-        expr: 'sum(rate(http_requests_total{job="observatorium-observatorium-api", namespace="$namespace", handler=~"$handler",code=~"5.."}[5m]))\n/\nsum(rate(http_requests_total{job="observatorium-observatorium-api", namespace="$namespace", handler=~"$handler"}[5m]))',
-        legendFormat: 'errors',
+        expr: 'sum by (code) (rate(http_requests_total{job="observatorium-observatorium-api", namespace="$namespace", handler=~"$handler",code=~"5.."}[5m]))\n/\nsum by (code) (rate(http_requests_total{job="observatorium-observatorium-api", namespace="$namespace", handler=~"$handler"}[5m]))',
```
douglascamata, line 338: The new query will always return `1` because it's doing the division only for the vectors with matching labels, so `code=500` on numerator will only be divided by `code=500` on denominator:

```
sum by (code) (rate(http_requests_total{job="observatorium-observatorium-api",handler=~"query|query_legacy",code=~"5.."}[5m]))
/
sum by (code) (rate(http_requests_total{job="observatorium-observatorium-api",handler=~"query|query_legacy"}[5m]))
```

I think you will have to do something like the query below. Note that the denominator of the division is explicitly transform into a scalar after summing the rate for all the matching vectors.

```
sum by (code) (rate(http_requests_total{job="observatorium-observatorium-api",handler=~"query|query_legacy",code=~"5.."}[5m]))
/
scalar(sum(rate(http_requests_total{job="observatorium-observatorium-api",handler=~"query|query_legacy"}[5m])))
```

douglascamata (changes requested):
I think the query isn't returning exactly what we expect now.

observability/dashboards/observatorium-api.libsonnet
jessicalins: good catch! thanks, fixed it

douglascamata (approved):
LGTM 👍

PR #338, opened 2022-09-29
Fix alertmanager discovery for logs ruler

periklis:
According to staging deployment the port to look for dns-based discovery is `_http` and not `_web`:
image


saswatamcode (approved):
LGTM!

PR #346, opened 2022-10-05
Disable compression on receive

matej-g:

Since `0.28.0`, receiver uses snappy compression for gRPC requests by default (https://github.com/thanos-io/thanos/pull/5575#discussion_r939525460). However, we noticed in our case this leads to an increased memory consumption on receiver replicas, especially in Telemeter where the requests are bigger (contain more series). Since we're better of with no compression / less memory overhead in our case, I'm disabling it.


PhilipGough:
I wonder do we want to make this a parameter with the default value of none for future proofing?

Saying that I'm ok with this directly for now

matej-g:
> I wonder do we want to make this a parameter with the default value of none for future proofing?
> 
> Saying that I'm ok with this directly for now

I think we can cross that bridge when we come to it :+1:

PR #349, opened 2022-10-06
Add template for Loki ruler CRDs

periklis:
The Loki Ruler CRDs are not by default present on our clusters. Thus this PR provides a template to include the CRDs for Loki AlertingRule and RecordingRule resources.


saswatamcode (approved):
Thanks! LGTM as far as I understand. 🙂

PR #356, opened 2022-10-11
Add querier with new PromQL engine

saswatamcode:
This PR adds a new `observatorium-volcano-query` deployment which deploys Thanos Querier with the new PromQL engine enabled. This also adds some new params to allow a different Thanos image to be used while deploying these new Queriers. This querier is not exposed to tenants.

I'm not mirroring traffic to this yet, as the new engine still doesn't support many functions, modifiers, or even functions within functions.
Mirroring traffic would lead to several "not implemented" errors right now, but can be enabled soon once some WIP stuff is merged upstream. For now, it would be valuable to process known queries and compare results with old Querier. 🙂

### DIFF
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -45,6 +45,7 @@ objects:
 - apiVersion: v1
   kind: ServiceAccount
   metadata:
+    annotations: {}
     labels:
       app.kubernetes.io/component: database-compactor
       app.kubernetes.io/instance: observatorium
@@ -749,6 +750,7 @@ objects:
 - apiVersion: v1
   kind: ServiceAccount
   metadata:
+    annotations: {}
     labels:
       app.kubernetes.io/component: rule-evaluation-engine
       app.kubernetes.io/instance: observatorium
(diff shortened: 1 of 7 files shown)


bwplotka (approved):
Amazing, thanks!

resources/services/observatorium-metrics-template.yaml
```diff
@@ -990,6 +992,188 @@ objects:
...
+  metadata:
+    labels:
+      app.kubernetes.io/component: query-layer
+      app.kubernetes.io/instance: observatorium
+      app.kubernetes.io/name: thanos-query
+      app.kubernetes.io/part-of: observatorium
+      app.kubernetes.io/version: ${THANOS_VOLCANO_IMAGE_TAG}
+    name: observatorium-volcano-query
+  spec:
+    replicas: 2
```
matej-g, line 1006: We could parametrize this as well, but probably for playing around 2 replicas are ample :+1:

matej-g (approved):
Looks good! Even though I'm not a big fan of temporary changes, given this is an important feature to test, I'm happy to have it in. We could think in the future about adding something like a canary deployment, that we could scale / configure to test experimental features more easily.

resources/services/observatorium-metrics-template.yaml
bwplotka: 👍🏽

fpetkovski:
@saswatamcode feel free to raise a separate PR with the improved fallback mechanism so that you can get unblocked on getting some real world traffic :)

PR #357, opened 2022-10-11
Enable hashing algorithm for receive to be set via parameter

PhilipGough:
Note, this is a breaking change which requires the use of Thanos >= v0.28.0

### DIFF
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -2144,6 +2144,7 @@ objects:
             "type": "JAEGER"
           - --receive.default-tenant-id=FB870BF3-9F3A-44FF-9BF7-D7A047A52F43
           - --receive.grpc-compression=none
+          - --receive.hashrings-algorithm=${THANOS_RECEIVE_HASHRINGS_ALGORITHM}
           env:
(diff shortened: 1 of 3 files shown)


resources/services/observatorium-metrics-template.yaml
```diff
@@ -4335,6 +4336,8 @@ parameters:
   value: 4d
 - name: THANOS_RECEIVE_HASHRING_SERVICE_NAME
   value: observatorium-thanos-receive-default
+- name: THANOS_RECEIVE_HASHRINGS_ALGORITHM
+  value: hashmod
```
matej-g, line 4340: Just thinking if we should make this `ketama`, since that is technically our default now everywhere?

PhilipGough: I'm easy. `hashmod` is the default in Thanos and we have explicitly set `ketama` as param values (which I prefer anyway personally), but I dont mind if we want to change this

matej-g: Cool, just two easy dudes not caring :sunglasses:. Let's keep it, it's a moot point.

matej-g (approved):

PR #361, opened 2022-10-11
Add SLO rules for Loki AlertingRule and RecordingRule CRs

periklis:


resources/services/observatorium-template.yaml
```diff
@@ -158,6 +158,67 @@ objects:
...
+      - alert: NeverFiringAlert
+        annotations:
+          description: Does not fire!
+          message: Alert not fired.
+        expr: 1 < 0
+        for: 1m
+        labels:
+          severity: page
+          source: alertingrule.loki.grafana.com
+    tenantID: rhobs
```
saswatamcode, line 199: I guess this `tenantID` is specific to Loki?

saswatamcode (approved):
LGTM!

One small question!

resources/services/observatorium-template.yaml
periklis: Yes this the only thing it deviates from PrometheusRule. In general we introduced only tenant-aware CRDs in the Loki Operator. This operator deploys Loki with multi-tenancy by default. Of course you can use a single fake ID for putting anything into a single spot

saswatamcode: Ack Thanks!

saswatamcode (approved):

PR #362, opened 2022-10-12
Update Hypershift RBAC to read only

ninabauer:
Updating hypershift-platform RBAC to read only to prevent unexpected writes.

OSD-13206


jessicalins (approved):
LGTM :)

PR #364, opened 2022-10-13
Bump Thanos Store liveness probe timeout

douglascamata:
This PR bumps the Thanos Store liveness probe timeout from the default 1s to a manually assigned 30s.

This helps liveness probes to fail when the Store is under heavy load.


PhilipGough (approved):
lgtm

PR #369, opened 2022-10-28
Add proxy strategy flag (k-way merge feature)

matej-g:

Add the hidden flag to make it possible to toggle the new k-way merge feature. The flag can be either `eager` (old behavior) or `lazy` (new k-way merge behavior).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,8 @@ validate: $(OC)
 
 .PHONE: sync-crds
 sync-crds: $(YQ) $(GOJQ)
-	@curl --no-progress-meter https://raw.githubusercontent.com/grafana/loki/main/operator/bundle/manifests/loki.grafana.com_alertingrules.yaml | $(YQ) eval -j > $(CRD_DIR)/loki.grafana.com_alertingrules.libsonnet
-	@curl --no-progress-meter https://raw.githubusercontent.com/grafana/loki/main/operator/bundle/manifests/loki.grafana.com_recordingrules.yaml | $(YQ) eval -j > $(CRD_DIR)/loki.grafana.com_recordingrules.libsonnet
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -42,8 +42,8 @@ validate: $(OC)
 
 .PHONE: sync-crds
 sync-crds: $(YQ) $(GOJQ)
-	@curl --no-progress-meter https://raw.githubusercontent.com/grafana/loki/main/operator/bundle/manifests/loki.grafana.com_alertingrules.yaml | $(YQ) eval -j > $(CRD_DIR)/loki.grafana.com_alertingrules.libsonnet
```
matej-g, line 45: I removed this flag because apparently it's only available on some newer versions of `curl`.

services/observatorium-metrics-template-overwrites.libsonnet
```diff
@@ -308,6 +308,14 @@ local thanosRuleSyncer = import './sidecars/thanos-rule-syncer.libsonnet';
           template+: {
             spec+: {
               securityContext: {},
+              containers: [
+                if c.name == 'thanos-query' then c {
+                  args+: [
+                    '--grpc.proxy-strategy=${THANOS_QUERIER_PROXY_STRATEGY}',
```
saswatamcode, line 314: Maybe we should add to kube-thanos first?

saswatamcode (approved):
LGTM!
Small idea, but can be done later as well.

services/observatorium-metrics-template-overwrites.libsonnet
matej-g: Not sure since it's hidden flag (does `kube-thanos` have hidden flags?), but once it's added we can obviously remove this override :+1:

PR #370, opened 2022-10-31
Add missing runbook link

saswatamcode:

### DIFF
--- a/docs/sop/observatorium.md
+++ b/docs/sop/observatorium.md
@@ -101,7 +101,10 @@
   - [ThanosRuleRuleEvaluationLatencyHigh](#thanosruleruleevaluationlatencyhigh)
   - [ThanosRuleTSDBNotIngestingSamples](#thanosruletsdbnotingestingsamples)
   - [ThanosRuleIsDown](#thanosruleisdown)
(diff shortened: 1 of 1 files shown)


jessicalins (approved):
thanks!

docs/sop/observatorium.md
```diff
@@ -1812,11 +1815,11 @@ Respective component that is supposed to be running is not running.
 
 ---
 
-## ObservatoriumAPIMetricsErrorsSLOBudgetBurn
+## ObservatoriumAPIMetricsErrorsSLOBudgetBurn5mand1h
```
matej-g, line 1818: Should we make this generic, something like `ObservatoriumAPIMetricsErrorsSLOBudgetBurnXmandYh`? I just want to avoid confusion with different `m` and `h` values between different alerts. So I would suggest either making this heading more generic or make separate section for each alert.

saswatamcode: Yeah I'll try to make the heading generic. Problem is the runbook links get autogenerated from jsonnet, so the heading need to be there is markdown

matej-g: Ah right, in my world I ignore jsonnet :joy:, I only reviewed the generated portion

PR #372, opened 2022-11-07
Sync telemeter rules to include new vCPU-hours rule

saswatamcode:
This updates telemeter lib to roll out new rule added in https://github.com/openshift/telemeter/pull/435


PhilipGough (approved):
lgtm

PR #374, opened 2022-11-16
MGDAPI-4934 : Adding metrics to match metrics added to CMO

Boomatang:
JIRA: https://issues.redhat.com/browse/MGDAPI-4934

This PR syncs metrics that have being merged to CMO. 

The CMO PR can be found here: https://github.com/openshift/cluster-monitoring-operator/pull/1771


moadz:
LGTM, sidenote: why are we using Telemeter here if they're RHOAM specific metrics?  I believe theres a RHOAM tenant in the MST instance...

moadz (approved):

Boomatang:
Thank you @moadz, who do we now get to merge this change. On the tenant in the MST instance I believe the reason is we are not aloud put the required secrets on to clusters that are controlled by the customer.

moadz:
@Boomatang makes sense! We need to do a `telemeter-stage` and `telemeter-production` rollout to propagate these changes. Will ping observatorium-support.

PR #382, opened 2022-12-12
Split hypershift RBAC into two tenants

saswatamcode:
This PR adds a new `hypershift-platform-staging` tenant in the rbac, and splits the rbac SAs across `hypershift-platform` and `hypershift-platform-staging` tenants. (would need to update saas-tenants)

Effectively,

`service-account-observatorium-hypershift-platform` can read/write to `hypershift-platform` &
`service-account-observatorium-hypershift-platform-staging` can read/write to `hypershift-platform-staging`

**Do not merge, until we have new tenant**

### DIFF
--- a/configuration/observatorium/rbac.go
+++ b/configuration/observatorium/rbac.go
@@ -335,7 +335,7 @@ func attachBinding(o *observatoriumRBAC, opts bindingOpts) {
 	var subs []rbac.Subject
 	for _, e := range opts.envs {
 		// observatorium-hypershift-platform-staging is the only tenant that does not
-		// follow conventions, due to the being present in a unique environment alongside
+		// follow conventions, due to them being present in an unique environment alongside
 		// their production tenant on rhobsp02ue1.
 		errMsg, ok := tenantNameFollowsConvention(opts.name)
 		if !ok && opts.name != "observatorium-hypershift-platform-staging" {
--- a/configuration/observatorium/rbac.go
+++ b/configuration/observatorium/rbac.go
@@ -214,12 +214,16 @@ func GenerateRBAC(gen *mimic.Generator) {
 	})
 
 	// hypershift staging
+	// observatorium-hypershift-platform-staging is the only tenant that does not
+	// follow conventions, due to them being present in an unique environment alongside
+	// their production tenant on rhobsp02ue1.
 	attachBinding(&obsRBAC, bindingOpts{
-		name:    "observatorium-hypershift-platform-staging",
-		tenant:  hypershiftStagingTenant,
-		signals: []signal{metricsSignal},
-		perms:   []rbac.Permission{rbac.Write, rbac.Read},
-		envs:    []env{productionEnv},
+		name:                "observatorium-hypershift-platform-staging",
+		tenant:              hypershiftStagingTenant,
+		signals:             []signal{metricsSignal},
+		perms:               []rbac.Permission{rbac.Write, rbac.Read},
+		envs:                []env{productionEnv},
+		skipConventionCheck: true,
 	})
 
 	// RHOBS Logs only tenants
@@ -275,11 +279,12 @@ type observatoriumRBAC struct {
 type bindingOpts struct {
 	// NOTE(bwplotka): Name is strongly correlated to subject name that corresponds to the service account username (it has to match it)/
 	// Any change, require changes on tenant side, so be careful.
-	name    string
-	tenant  tenantID
-	signals []signal
-	perms   []rbac.Permission
-	envs    []env
+	name                string
+	tenant              tenantID
+	signals             []signal
+	perms               []rbac.Permission
+	envs                []env
+	skipConventionCheck bool
 }
 
 func getOrCreateRoleName(o *observatoriumRBAC, tenant tenantID, s signal, p rbac.Permission) string {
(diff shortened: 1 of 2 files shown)


configuration/observatorium/rbac.go
```diff
@@ -306,18 +316,11 @@ func attachBinding(o *observatoriumRBAC, opts bindingOpts) {
...
-			err := fmt.Sprintf(
-				"found name breaking conventions with environment suffix: %s, should be: %s",
-				opts.name,
-				strings.TrimRight(strings.TrimSuffix(opts.name, string(e)), "-"),
-			)
-			mimic.Panicf(err)
-		}
-		n := fmt.Sprintf("service-account-%s-%s", opts.name, e)
-		if e == productionEnv {
-			n = fmt.Sprintf("service-account-%s", opts.name)
```
douglascamata, line 319: Why is this being changed? Why are we removing the conventions check?

configuration/observatorium/rbac.go
```diff
@@ -209,7 +210,16 @@ func GenerateRBAC(gen *mimic.Generator) {
...
+		envs:    []env{productionEnv},
+	})
+
+	// hypershift staging
+	attachBinding(&obsRBAC, bindingOpts{
+		name:    "observatorium-hypershift-platform-staging",
+		tenant:  hypershiftStagingTenant,
+		signals: []signal{metricsSignal},
+		perms:   []rbac.Permission{rbac.Write, rbac.Read},
+		envs:    []env{stagingEnv},
```
douglascamata, line 222: I wonder which environment should be there. It's a staging account from the tenant point of view, but from RHOBS point of view data goes to production.

wdyt, @saswatamcode?

configuration/observatorium/rbac.go
saswatamcode: Because with the `hypershift-platform-staging` tenant, we effectively break conventions. I removed it, as otherwise working around it leads to a different SA name being generated.  😅

configuration/observatorium/rbac.go
saswatamcode: Yeah, the env here is actually appended to SA name. I would like it to be prod here, but with this new tenant, we effectively break conventions.

configuration/observatorium/rbac.go
douglascamata: Do we have any documentation or good reason for breaking the convention? IMHO we should fix the service account. The convention was put there for a reason and this check already detected issues in the past.

configuration/observatorium/rbac.go
douglascamata: Hummm, I see. Can we documented then that the environment field is from the point of view of the tenant and not from which RHOBS environment the data goes to?

configuration/observatorium/rbac.go
douglascamata: We can also improve the code to allow exceptions and skip the check. But let's not open the exception door to everyone.

This also relates to my other comment. If we leave the service account name there with the `-staging` suffix and set `[]env{productionEnv}` I think it'll work.

configuration/observatorium/rbac.go
douglascamata: See a reply in my other comment that is somehow related to this.

configuration/observatorium/rbac.go
douglascamata: The key for this it work will be that if the env is `productionEnv`, we don't include any env in the subject name.

douglascamata: I tried with `productionEnv` for the HyperShift tenant and the configuration I got out of it makes sense. In the end, the `envs` mean RHOBS envs and not the tenant's.

This is the role binding:

```json
    {
      "name": "observatorium-hypershift-platform-staging",
      "subjects": [
        {
          "name": "service-account-observatorium-hypershift-platform-staging",
          "kind": "user"
        }
      ],
      "roles": [
        "hypershift-platform-staging-metrics-write",
        "hypershift-platform-staging-metrics-read"
      ]
    },
```

And these are the roles:

```json
    {
      "name": "hypershift-platform-staging-metrics-write",
      "resources": [
        "metrics"
      ],
      "tenants": [
        "hypershift-platform-staging"
      ],
      "permissions": [
        "write"
      ]
    },
    {
      "name": "hypershift-platform-staging-metrics-read",
      "resources": [
        "metrics"
      ],
      "tenants": [
        "hypershift-platform-staging"
      ],
      "permissions": [
        "read"
      ]
    }
```

saswatamcode: You're right, `productionEnv` does work, I missed this. Seems like we can add the check back in as well. It doesn't check prefix in all envs but only for `opts.Env`

configuration/observatorium/rbac.go
```diff
@@ -306,18 +334,19 @@ func attachBinding(o *observatoriumRBAC, opts bindingOpts) {
...
-				opts.name,
-				strings.TrimRight(strings.TrimSuffix(opts.name, string(e)), "-"),
-			)
-			mimic.Panicf(err)
+		// observatorium-hypershift-platform-staging is the only tenant that does not
+		// follow conventions, due to them being present in an unique environment alongside
+		// their production tenant on rhobsp02ue1.
+		errMsg, ok := tenantNameFollowsConvention(opts.name)
+		if !ok && opts.name != "observatorium-hypershift-platform-staging" {
+			mimic.Panicf(errMsg)
```
douglascamata, line 342: Instead of having the data to decide the skip hardcoded in the convention check, what if we created a new field in the `bindingOpts` struct like `skipsConventionCheck bool`?

configuration/observatorium/rbac.go
douglascamata: Almost LGTM, see https://github.com/rhobs/configuration/pull/382#discussion_r1053164673. 😄

configuration/observatorium/rbac.go
saswatamcode: Implemented in https://github.com/rhobs/configuration/pull/382/commits/e54920cee9294019b83d78bee588b362792ff8f7

douglascamata (approved):
LGTM :shipit:

PR #386, opened 2023-01-13
Setting all availability/latency SLO dashboard rows to instant query …

moadz:
…instead of range (as the SLO itself is a range query)

Noticed some SLO dashboard rows weren't loading after Grafana was moved into VPN. Subsequently noticed we were 'double ranging' (query_range on range vector queries that return a single value). 

This change means each visualised SLO is only queried for its most recent value. 

> Future enhancements
* `$$time_range` should be parameterised into all the range vector SLO queries, that way the set time range will give you the SLO for that time period.


observability/dashboards/slo.libsonnet
```diff
@@ -311,6 +313,8 @@ function(instanceName, environment, dashboardName) {
           hide: false,
           interval: '',
           legendFormat: '',
+          instant: true,
+          range: false,
```
moadz, line 317: yeah the double declaration is really stupid, but it's probably for compatibility.

saswatamcode (approved):
LGTM!

PR #388, opened 2023-01-17
Use OpenShift template var to select query engine

douglascamata:

The default is the `prometheus` PromQL engine, to preserve backwards compatibility. 

This will allow us to start testing with only running the new PromQL engine by setting it to `thanos`.


saswatamcode (approved):
LGTM!

PhilipGough (approved):

PR #398, opened 2023-02-01
Add missing UIDs for rhobs logs dashboards

periklis:
Some of the mixin-vendored Dashboard from grafana/loki are missing UIDs. This is not a mandatory requirement for Grafana dashboard being provisioned though a ConfigMap. This change makes the `observatorium-logs/grafana-dashboards-template.yaml` deployable again.


saswatamcode (approved):
LGTM!

PR #401, opened 2023-02-07
Add Thanos Store series touched limit configuration

douglascamata:
Tried to use the nice jsonnet function constructor from https://github.com/thanos-io/kube-thanos/commit/d2cac7b14a147e4c79edd6f02fba7afeafbe9702, but it doesn't work because we use OpenShift Templates parameters to configure most of the things, including limits, and the jsonnet function requires the limits values to be a number.
Because of this, the implementation follows the pattern already set by other components, like the Thanos Query overwrites and others in the same file.

The default value is `0` (zero), which means "no limit".

### DIFF
--- a/jsonnetfile.lock.json
+++ b/jsonnetfile.lock.json
@@ -336,8 +336,8 @@
           "subdir": "jsonnet/kube-thanos"
         }
       },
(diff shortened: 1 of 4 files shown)


moadz (approved):
LGTM

PR #409, opened 2023-02-15
[rhobs-logs] Refactor Loki dashboards

periklis:
The following change set refactors the Loki dashboards addressing the following:
1. Make expression replacements local to the panels and dashboard they apply. This enables removing them easier when we update the loki-mixin.
2. Remove obsolete dashboard rows, e.g. related to Big Table, Dynamo, GCS, Cassandra, because we run only Loki on boltdb-shipper with S3.
3. Replace cortex gateway references with `API` that is the official name of our RHOBS API in other dashboards.
4. Use only RHOBS cluster names for metrics datasources.
5. Provide a template parameters for available namespaces as a dashboard variable.
6. Drop `loki-logs` dashboard as it does not work with our RHOBS logs label selectors.

### Hints for the reviewer to test the boards:
```sh
$ oc process --local -f ./resources/observability/grafana/observatorium-logs/grafana-dashboards-template.yaml | jq -r '.items[].data' | jq -r '."loki-operational.json"' > loki-operational.json
$ oc process --local -f ./resources/observability/grafana/observatorium-logs/grafana-dashboards-template.yaml | jq -r '.items[].data' | jq -r '."loki-chunks.json"' > loki-chunks.json
$ oc process --local -f ./resources/observability/grafana/observatorium-logs/grafana-dashboards-template.yaml | jq -r '.items[].data' | jq -r '."loki-reads.json"' > loki-reads.json
$ oc process --local -f ./resources/observability/grafana/observatorium-logs/grafana-dashboards-template.yaml | jq -r '.items[].data' | jq -r '."loki-reads.json"' > loki-reads.json
```

PR #411, opened 2023-02-17
[rhobs-logs] Load namespace variable from selected datasource

periklis:
Follow up on #409 that makes loading the namespace variable based on the selected data-source. This allows inspecting all clusters a lot easier.


saswatamcode (approved):
LGTM!

PR #415, opened 2023-02-23
[rhobs-logs] Refactor dashboard customizations

periklis:
Refactor long due issues in managing dashboard customizations:
1. Replace for-comprehensions in patching panels with more functional alternatives.
2. Add units to all latency graphs.

Depends on 
- #414

### DIFF
--- a/observability/config.libsonnet
+++ b/observability/config.libsonnet
@@ -33,41 +33,111 @@ local thanos = (import '../services/observatorium-metrics.libsonnet').thanos;
     },
   },
 
-  local withLokiMetricsDatasource = function(ds, key) ds + (
-    if ds.name == key then {
-      regex: '${OBSERVATORIUM_DATASOURCE_REGEX}',
-      current: {
-        selected: true,
-        text: '${OBSERVATORIUM_API_DATASOURCE}',
-        value: '${OBSERVATORIUM_API_DATASOURCE}',
-      },
-    } else {}
-  ),
-
-  local withLokiMetricsNamespace = function(ns, ds, value) ns + (
-    if ns.label == value then {
-      datasource: {
-        type: 'prometheus',
-        uid: '${%s}' % ds,
-      },
-      current: {
-        selected: true,
-        text: '${OBSERVATORIUM_API_NAMESPACE}',
-        value: '${OBSERVATORIUM_API_NAMESPACE}',
-      },
-      definition: 'label_values(kube_pod_info, namespace)',
-      query: {
-        query: 'label_values(kube_pod_info, namespace)',
-        refId: 'StandardVariableQuery',
-      },
-      regex: 'observatorium-logs|mst-.+',
-    } else {}
-  ),
-
-  local defaultLokiTags = function(t)
-    std.uniq(t + ['observatorium', 'observatorium-logs']),
-
   loki: {
+    local withDatasource = function(ds) ds + (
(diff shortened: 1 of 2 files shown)


observability/config.libsonnet
```diff
@@ -126,76 +173,9 @@ local thanos = (import '../services/observatorium-metrics.libsonnet').thanos;
...
-                    'kube_pod_container_status_restarts_total{container=~"observatorium-loki.+", ',
-                  ),
-              }
-              for t in p.targets
-            ],
-          }
-          for p in super.panels
-          if !std.member(['Consul', 'Big Table', 'GCS', 'Dynamo', 'Cassandra', 'Azure Blob'], p.title)
-        ],
+        panels: mapPanels([replaceMatchers, withLatencyAxis], dropPanels(super.panels)),
```
saswatamcode, line 176: This line does quite a lot and isn't super clear. Maybe we need to have some comments explaining entire functionality here?

observability/config.libsonnet
```diff
@@ -33,41 +33,99 @@ local thanos = (import '../services/observatorium-metrics.libsonnet').thanos;
...
+
+    local mapPanels = function(funcs, panels)
+      [
+        std.foldl(function(agg, fn) fn(agg), funcs, p) + (
+          if std.objectHas(p, 'panels') then {
+            panels: mapPanels(funcs, p.panels),
+          } else {}
+        )
+        for p in panels
+      ],
```
saswatamcode, line 120: I might be a  bit biased with Go, so feel free to disagree with me here, but feels like this function is quite complex and not really readable without comments explaining what it is doing. 

The `std.foldl` alone is difficult to reason about with the recursion. And with a lack of logging/print, understanding and modifying this, if there are some potential changes needed in the future, might prove to be difficult.

observability/config.libsonnet
```diff
@@ -33,41 +33,99 @@ local thanos = (import '../services/observatorium-metrics.libsonnet').thanos;
...
+          // TODO: This substitution is needed because the 'replaceClusterMatchers' function in upstream lib 'loki-operational.libsonnet' misses 'cluster=~"$cluster"' substitution.
+          // Remove this substitution when it has been added to upstream.
+          { from: 'cluster=~"$cluster",', to: '' },
+          { from: 'job=~"$namespace/cortex-gw(-internal)?",', to: '' },
+          { from: 'kube_pod_container_status_restarts_total{ ', to: 'kube_pod_container_status_restarts_total{container=~"observatorium-loki.+", ' },
+          { from: ' * 1e3', to: '' },
+          { from: 'pod=~"distributor.*"', to: 'pod=~".*distributor.*"' },
+          { from: 'pod=~"ingester.*"', to: 'pod=~".*ingester.*"' },
+          { from: 'pod=~"querier.*"', to: 'pod=~".*querier.*"' },
+          { from: 'job=~"$namespace/ingester",', to: 'job="observatorium-loki-ingester-http"' },
```
saswatamcode, line 94: I really like this way of replacing the matchers, looks much cleaner! I'm usually a bit worried about modifying regex matchers in PromQL in jsonnet, but this seems like simple replace. 🙂

saswatamcode (commented):
Thanks! This looks good, mod some comments about the more complex functions.

saswatamcode (approved):
LGTM!

PR #417, opened 2023-02-28
[rhobs-logs] Fix remaining issues after loki-mixin update

periklis:
Fixes provided:
- Fix syntax error in chunk reason panel expression
- Remove unnecessary panels for zone-aware data replication because we do not make use of it.
- Add Loki retention dashboard to have an oversight of compacting index files.


saswatamcode (approved):
LGTM!

PR #421, opened 2023-03-09
Implement CI using CircleCI

vprashar2929:
This PR is in continuation to [this](https://github.com/rhobs/configuration/pull/418). It contains implementing RHOBS configuration CI using CircleCI and addresses some of the drawbacks of using GitHub Actions.
Currently the CI pipeline performs following actions:
1. Run microshift on Linux VM that is provided by CircleCI. Since RHOBS configs are specific to OCP so it's best to use OCP env to deploy and test these instead of kind/minikube. 
2. Deploy's the generated manifests on microshift cluster with minimum resources. 
3. Run a small test using observatorium/up (Like in [observatorium e2e.sh](https://github.com/observatorium/observatorium/blob/main/configuration/tests/e2e.sh)) to verify send/receive metrics.

Using circleci we are getting good [resources](https://circleci.com/docs/configuration-reference/#linuxvm-execution-environment) out of the box than existing GitHub Actions runners.


simonpasquier:
I'm really an outsider here but I agree with @douglascamata that more context would be helpful
 I'd keep also in mind that CircleCI might reduce the free plan's resources at any time without prior notice. Do we have a backup plan for this possibility?

vprashar2929:
> hey @vprashar2929, can you mention the drawbacks of Github Actions that makes us fallback to Circle CI, please, as we spoke some time ago? It's valuable to have the reasoning of the decision for documentation purposes.

So the only difference between the two according to our CI use case is in terms of out of the box (free of cost) resources which both has to offer. GitHub Actions provide [2 vCPU and 7GB or ram](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) which are not sufficient enough to deploy complete RHOBS components in one go. Whereas CircleCI provides [4 vCPU and 15GB of ram](https://circleci.com/docs/using-linuxvm/#available-linuxvm-resource-classes). 
Since we want to deploy RHOBS with minimum resources and run a simple test on generated manifests than currently it is easy to achieve using CircleCI instead of GitHub Actions.

vprashar2929:
> I'm really an outsider here but I agree with @douglascamata that more context would be helpful I'd keep also in mind that CircleCI might reduce the free plan's resources at any time without prior notice. Do we have a backup plan for this possibility?

Oh reduce free plan resources hurts 😅. Well in that case what we can do is deploy each components individually and check its rollout status. But in this case we won't be able to deploy all components at once and run tests on it. It can only validate the deployments generated manifests.

douglascamata:
@simonpasquier @vprashar2929 my proposals for the case where CircleCI reduces the size of the free runner:

1. We join the paid plan (lol) of whichever of the two vendors has the best cost/benefit ratio for our case.
2. We run a private Github Action Runner for the builds of the RHOBS org.

I personally don't worry too much in case we need a backup plan. 👍

vprashar2929:
@douglascamata Absolutely agree with you. We can definitely do a cost analysis between the two and choose what best suits us according to cost. Provided we are willing to pay the bucks😄
Also I am open to self-hosted GitHub runners specifically for our RHOBS repo as in that case we can have a dedicated OCP cluster running and escape from resources crunch completely.

tests/ci/ci_test.sh
douglascamata, line 1: @vprashar2929 was wondering if we could better separate here the deploy procedures and the test ones, so that we can reuse the test part in a post-deploy job. What do you think? This is not a blocker though. I am happy with merging this as is now and then follow up with the separation and more tests.

douglascamata (approved):
LGTM. 

Would be great to have a review from other team members (cc @rhobs/team-monitoring)

tests/ci/ci_test.sh
```diff
@@ -0,0 +1,140 @@
+#!/bin/bash
+
+set -e
```
sthaha, line 3: I think it s best to have `-u` set for all bash script. I good template to follow in imho is 
```bash

#!/usr/bin/env bash

set -eu -opipefail

main() {
}

main "$@"
```

tests/ci/ci_test.sh
```diff
@@ -0,0 +1,140 @@
...
+set -o pipefail
+
+ARTIFACT_DIR="${ARTIFACT_DIR:-/tmp/artifacts}"
```
sthaha, line 6: declare -r for all readonly vars

tests/ci/ci_test.sh
```diff
@@ -0,0 +1,140 @@
...
+INFO="INFO"
```
sthaha, line 7: why is this required? where do we change this variable?

tests/ci/ci_test.sh
```diff
@@ -0,0 +1,140 @@
...
+log_info(){
+    echo "$(date "+%Y-%m-%d %H:%M:%S") [$INFO] $1"
+}
+check_status(){
+    res=$1
+    namespace=$2
+    log_info "Checking status of $res inside $namespace"
+    oc rollout status $res -n $namespace --timeout=5m || (must_gather "$ARTIFACT_DIR" && exit 1)
```
sthaha, line 15: Do we need a subshell here? 
```suggestion
    oc rollout status $res -n $namespace --timeout=5m ||  { 
        must_gather "$ARTIFACT_DIR"
        return 1
      }
```

tests/ci/ci_test.sh
```diff
@@ -0,0 +1,140 @@
...
+}
+prereq(){
+    log_info "Deploying prerequisites on cluster"
+    oc apply -f pre-requisites.yaml
+    oc create -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/master/bundle.yaml 
+    oc process -f ../../resources/crds/observatorium-logs-crds-template.yaml | oc apply -f -
+
+}
+minio(){
```
sthaha, line 24: nit: a blank line between functions will help with readability.

tests/ci/ci_test.sh
```diff
@@ -0,0 +1,140 @@
...
+}
+check_status(){
+    res=$1
+    namespace=$2
+    log_info "Checking status of $res inside $namespace"
+    oc rollout status $res -n $namespace --timeout=5m || (must_gather "$ARTIFACT_DIR" && exit 1)
+}
+prereq(){
+    log_info "Deploying prerequisites on cluster"
+    oc apply -f pre-requisites.yaml
```
sthaha, line 19: should we use `apply` ? or `create` ?

tests/ci/ci_test.sh
```diff
@@ -0,0 +1,140 @@
...
+    oc create -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/master/bundle.yaml 
+    oc process -f ../../resources/crds/observatorium-logs-crds-template.yaml | oc apply -f -
+
+}
+minio(){
+    log_info "Deploying resources inside minio namespace"
+    oc create ns minio || true
+    sleep 5
```
sthaha, line 27: sleep is usually a code smell. It is better to "poll" or wait for something to happen.

tests/ci/ci_test.sh
```diff
@@ -0,0 +1,140 @@
...
+    oc process -f ../minio-template.yaml -p MINIO_CPU_REQUEST=30m -p MINIO_CPU_LIMITS=50m -p MINIO_MEMORY_REQUEST=50Mi -p MINIO_MEMORY_LIMITS=100Mi --local -o yaml | sed -e 's/storage: [0-9].Gi/storage: 0.25Gi/g' | oc apply -n minio -f -
```
sthaha, line 28:
```suggestion
    oc process -f ../minio-template.yaml \
      -p MINIO_CPU_REQUEST=30m \
      -p MINIO_CPU_LIMITS=50m \
      -p MINIO_MEMORY_REQUEST=50Mi \
      -p MINIO_MEMORY_LIMITS=100Mi \
      --local -o yaml | 
    sed -e 's/storage: [0-9].Gi/storage: 0.25Gi/g' | 
    oc apply -n minio -f -
```

tests/ci/ci_test.sh
```diff
@@ -0,0 +1,140 @@
...
+logs)
+    #TODO
+    ;;
+traces)
+    #TODO
+    ;;
+*)
+    echo "usage: $(basename "$0") { metrics | logs | traces }"
+    ;;
+esac
```
sthaha, line 140:
```suggestion
main() {
case $1 in
metrics)
    prereq
    minio
    dex
    observatorium_metrics
    observatorium
    run_test
    telemeter
    ;;
logs)
    #TODO
    ;;
traces)
    #TODO
    ;;
*)
    echo "usage: $(basename "$0") { metrics | logs | traces }"
    return 1
    ;;
esac
}

main "$@"
```

sthaha: Also instead of case, you can directly jump to a function like so 


```bash

ci.metrics() {
	:
}
ci.logs() {
	:
}

is_function() {
	local fn=$1
	shift
	[[ $(type -t "$fn") == "function" ]]
	return $?
}

main() {
	local fn=${1:-''}
	local ci_fn="ci.$fn"
	if ! is_function "$ci_fn"; then
		ci.help 1
	fi
	# call the function passing in the reset of the args
	shift
	$ci_fn "$@"
	return $?
}
```

tests/ci/ci_test.sh
vprashar2929: Definitely will add `-u` flag. Will avoid any undefined behaviour in script

tests/ci/ci_test.sh
vprashar2929: Earlier I used two variables for Logging purposes INFO and ERROR. But since we are gathering Artifacts in case of any failure I think its best to remove this

tests/ci/ci_test.sh
vprashar2929: Not mandatory. We can use `{}` along with `||` as well as it will also allow executing multiple commands if the `oc rollout status` fails

tests/ci/ci_test.sh
vprashar2929: Sure will add `oc wait` to wait for namespace to be created.

tests/ci/ci_test.sh
vprashar2929: VIM effects😅

tests/ci/ci_test.sh
vprashar2929: Awesome @sthaha 🙇🏼‍♂️. I haven't tried like this but this looks good. Will definitely use this!

PR #425, opened 2023-03-18
Update alerting and recording rules from v1beta1 to v1

vprashar2929:
This PR updates the alerting and recording rules from v1beta1 to v1

### DIFF
--- a/resources/services/observatorium-template.yaml
+++ b/resources/services/observatorium-template.yaml
@@ -158,7 +158,7 @@ objects:
               memory: ${MEMCACHED_EXPORTER_MEMORY_REQUEST}
         securityContext: {}
         serviceAccountName: observatorium-api-cache-memcached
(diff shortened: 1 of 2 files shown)


periklis (changes requested):
I would put this on hold for a moment. The main tool using these CRDS is not setup to read v1 AlertingRules yet namely `obsctl-reloader`:
https://github.com/rhobs/obsctl-reloader/blob/main/main.go#L16

vprashar2929:
Since the CRD's were now using v1 instead of v1beta1 I thought of bumping up in template as well. Last week I tried on enabling `obsctl-reloader` to read v1 rules but I wasn't able to do so. Got struck in versioning issue around `github.com/grafana/loki/operator` and its dependencies . Maybe lack of context around it😅. But anyways its better to hold it off

periklis (approved):

PR #428, opened 2023-03-20
receiver: Explicitly set the refresh interval to watch for changes to …

PhilipGough:
…hashring config

I don't expect any/much impact with this but would like to roll it out in isolation just to be sure. It appears looking from the Thanos codebase that https://github.com/fsnotify/fsnotify will look for FS changes and update but the fallback of 5m in case of a missed event is too large IMHO

```
     --receive.hashrings-file-refresh-interval=5m
                                 Refresh interval to re-read the hashring
                                 configuration file. (used as a fallback)
```


douglascamata (approved):
LGTM

PR #432, opened 2023-03-21
Reverts changes to GODEBUG env for http2

PhilipGough:
Reverts #430 and #429 

#429 proved effective at removing the culprit log  but we need to figure out if it makes sense to do so
```
caller="http2: server: error reading preface from client x.x.x.x" msg="read tcp x.x.x.x:8443->x.x.x.x:50454: read: connection reset by peer"18
```


douglascamata (approved):
LGTM

PR #433, opened 2023-03-24
[WIP]deploy loki via template in launch.sh

anpingli:


anpingli:
@vprashar2929  PTAL. 
By the ways, Loki pod can not be started due to resource limitation on my OCP cluster. I will have another try on the large cluster.

tests/launch.sh
```diff
@@ -41,31 +105,69 @@ telemeter() {
...
 }
-loki_crds(){
-    oc process -f ../resources/crds/observatorium-logs-crds-template.yaml | oc apply -f -
+
+lokistack(){
+    echo "deployLoki: in ${loki_namespace}"
+    cluster_name=$(oc config view -o jsonpath='{range .clusters[*]}{"\n"}{.name}' |egrep -v ':|^$')
+    app_domain=$(oc get ingresses.config/cluster -o jsonpath={.spec.domain})
+    default_storage_class=$(oc get sc |grep '(default)' |awk '{print $1}')
+    alertmanger_url="observatorium-alertmanager-mst.${app_domain}"
```
anpingli, line 114: What is the alertmanger_url if the alertmanger is deployed via launch.sh

vprashar2929:
> @vprashar2929 PTAL. By the ways, Loki pod can not be started due to resource limitation on my OCP cluster. I will have another try on the large cluster.

@anpingli Quick question is it mandatory to deploy Loki stack with resources provided in `observatorium-logs-template.yaml` I know they are big numbers but can't we reduce them to lower side. For eg. `observatorium-metrics-template.yaml` contains huge numbers but for deploying the components on testing OCP cluster we change those resources to minimum required and define them in `observatorium-metrics.test.env`. Can we do the same with Loki?

tests/launch.sh
vprashar2929: So alert manager deployed as part of `observatorium-metrics-template.yaml`. the url would be `observatorium-alertmanager.observatorium-metrics.svc.cluster.local`

tests/launch.sh
```diff
@@ -5,6 +5,70 @@
 set -e
 set -o pipefail
 
+set_aws_credentials()
```
vprashar2929, line 8: Setting credentials and creating s3 bucket can be handled inside `minio-template.yaml`. For testing purposes in case of metrics also we use minio. You can refer to minio template for setting s3 for loki

tests/launch.sh
```diff
@@ -5,6 +5,70 @@
...
+        aws s3api create-bucket --bucket $bucket_name --region $AWS_DEFAULT_REGION --create-bucket-configuration LocationConstraint=$AWS_DEFAULT_REGION
+        if [[ $? == 0 ]]  ;then
+           echo "CreateS3Bucket: created new bucket $bucket_name"
+        else
+            echo "CreateS3Bucket: Warning: !!!! create bucket $bucket_name failed, please create the bucket ${bucket_name} manually"
+        fi
+    fi
+}
+
+create_loki_secret() {
```
vprashar2929, line 57: We can create secrets using templates like we create for metrics in `observatorium-metrics-thanos-objectstorage-secret-template.yaml`

tests/launch.sh
```diff
@@ -41,31 +105,69 @@ telemeter() {
     oc apply --namespace telemeter -f telemeter-token-refersher-oidc-secret.yaml
     oc process --param-file=telemeter.test.env -f ../resources/services/telemeter-template.yaml | oc apply --namespace telemeter -f -
 }
-loki_crds(){
-    oc process -f ../resources/crds/observatorium-logs-crds-template.yaml | oc apply -f -
+
+lokistack(){
+    echo "deployLoki: in ${loki_namespace}"
+    cluster_name=$(oc config view -o jsonpath='{range .clusters[*]}{"\n"}{.name}' |egrep -v ':|^$')
```
vprashar2929, line 111: AFAIK the `observatorium-logs-template.yaml` contains all the necessary parameters like Namespace etc. You can create a separate env file like `observatorium.test.env` and can change those parameters according to the test OCP cluster like `LOKI_S3_SECRET` or `STORAGE_CLASS`. This way you dont have to define variables inside the `launch.sh`

PR #447, opened 2023-04-06
[rhobs-logs] Enhance http write timeouts and log level settings.

periklis:
1. Aligns the http write timeout with the default of the API (See observatorium/api#493)
2. Adds template parameter to set the loki log level per environment.


PhilipGough (approved):
lgtm

PR #454, opened 2023-04-12
Filter datasources on Thanos mixin dashboards

jacobbaungard:
With this commit we filter the available datasources on the grafana dashboards provided by the bundled thanos mixin dashboards. This should make it significantly easier to find the data needed.

The regex used currently will show the following datasources:
- app-sre-stage-01-prometheus
- rhobs-testing-cluster-prometheus
- rhobs-testing-prometheus
- rhobsp02ue1-cluster-prometheus
- rhobsp02ue1-prometheus
- telemeter-prod-01-prometheus

Any additional datasources starting with `rhobs` will be added if they become available.

Updates Thanos mixin to the latest version, which has the following PR as diff: https://github.com/thanos-io/thanos/pull/6273

### DIFF
--- a/jsonnetfile.lock.json
+++ b/jsonnetfile.lock.json
@@ -356,8 +356,8 @@
           "subdir": "mixin"
         }
       },
-      "version": "96fd6db6942290e1e520d556eed2e89df4e29eea",
-      "sum": "YWi7WSDfTecWr5xnXODfyxGcZUnXSw9Ba6lEanaqLW4=",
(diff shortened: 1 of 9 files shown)


jacobbaungard:
I wasn't able to figure out how to get rid of the `cluster-prometheus` datasources (which are not interesting for us), while still having a dynamic filter on `rhobs`. If someone is a regex wizard, feel free to add a suggestion :slightly_smiling_face:

PR #455, opened 2023-04-13
Filter datasources on Thanos / Receive Controller

jacobbaungard:
Following on from: https://github.com/rhobs/configuration/pull/454 we now use the same filter for the receive controller dashboard, which comes from the thanos-receive-controller rather than Thanos itself.

Pulls in the latest mixin from thanos-recieve-controller following: https://github.com/observatorium/thanos-receive-controller/pull/115


douglascamata (approved):
LGTM

PR #457, opened 2023-04-14
Filter namespaces on Receive Controller dashboard

jacobbaungard:
With this commit we filter the namespaces on the Thanos / Receive Controller dashboard, similar to the rest of the Thanos dashboards.

### DIFF
--- a/observability/config.libsonnet
+++ b/observability/config.libsonnet
@@ -8,6 +8,10 @@ local thanos = (import '../services/observatorium-metrics.libsonnet').thanos;
     targetGroups+:: {
       namespace: 'thanos_status',
     },
+    // Filter the namespaces in thanos-recieve-controller dashboard
+    hierarcies+:: {
+      namespace: 'thanos_status',
+    },
     overview+:: {
       title: '%(prefix)sOverview' % t.dashboard.prefix,
     },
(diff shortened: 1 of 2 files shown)


observability/config.libsonnet
```diff
@@ -8,6 +8,10 @@ local thanos = (import '../services/observatorium-metrics.libsonnet').thanos;
     targetGroups+:: {
       namespace: 'thanos_status',
     },
+    // Filter the namespaces in thanos-recieve-controller dashboard
+    hierarcies+:: {
```
moadz, line 12:
```suggestion
    hierarchies+:: {
```
But i'm assuming this typo comes from the mixin 🤣

moadz (approved):
LGTM :)

observability/config.libsonnet
jacobbaungard: Indeed: https://github.com/observatorium/thanos-receive-controller/blob/1a5357d4bd43d4291a0506ecaf09854ecb5ca6e6/jsonnet/thanos-receive-controller-mixin/config.libsonnet#L6

I guess we could fix that first though?

jacobbaungard: Attempted to fix the typo upstream while keeping backwards compatibility, but haven't managed to find a good solution yet. No reason for this to be stuck (imo), so I'll merge this for now.

PR #462, opened 2023-04-21
Add support to validate Logs on CI

vprashar2929:
This PR contains:

- Support for validating Logs template on CI
- Run `observatorium/up` on deployed Loki stack to do regression test
- Support for validating remaining components template like `jaeger,metric-federation-rule and parca`
- Fixes deployment of `Minio` and `Dex` on CI
- Some minor fixes/enhancements


services/observatorium-logs.libsonnet
```diff
@@ -270,6 +270,11 @@ local lokiCaches = (import 'components/loki-caches.libsonnet');
       tracing: {
         enabled: true,
       },
+      querier+: {
+        engine+: {
+          max_look_back_period: '5m',
+        },
+      },
     },
```
periklis, line 278: Why is this needed?

tests/ci/observatorium-logs.test.ci.env
```diff
@@ -0,0 +1,77 @@
+ALERTMANAGER_NAMESPACE=observatorium-metrics
+ALERTMANAGER_EXTERNAL_URL=my-route-observatorium-metrics.apps.vprashar-194066.qe.devcluster.openshift.com
```
periklis, line 2: I think this value can be something like `alertmanager.foobar.com`

tests/ci/observatorium-logs.test.ci.env
```diff
@@ -0,0 +1,77 @@
...
+RULES_OBJSTORE_S3_SECRET=rules-objstore-s3
+STORAGE_CLASS=kubevirt-hostpath-provisioner
```
periklis, line 4: Is the right storage class? If yes, can you please elaborate the setup?

periklis (approved):

services/observatorium-logs.libsonnet
vprashar2929: When I was trying to read/write logs after the Loki stack deployment the `observatorium/up` was able to write the logs but couldn't able to read them as the querier returned 0 matching results. I did some exploration around [querier config ](https://grafana.com/docs/loki/latest/configuration/#querier)and came across `max_look_back_period`. Since we deploy fresh Loki stack every time on CI and we do instant write/read so it made sense to include this param as `observatorium/up` was able to read the logs successfully.

tests/ci/observatorium-logs.test.ci.env
vprashar2929: Yes absolutely  will make this change

tests/ci/observatorium-logs.test.ci.env
vprashar2929: So when microshift container is deployed the default storage class which we get out of the box has name `kubevirt-hostpath-provisioner`
```
apiVersion: v1
items:
- apiVersion: storage.k8s.io/v1
  kind: StorageClass
  metadata:
    annotations:
      storageclass.kubernetes.io/is-default-class: "true"
    creationTimestamp: "2023-04-24T05:14:29Z"
    name: kubevirt-hostpath-provisioner
    resourceVersion: "989"
    selfLink: /apis/storage.k8s.io/v1/storageclasses/kubevirt-hostpath-provisioner
    uid: de28d577-f467-468e-bfde-5547c6000aa9
  provisioner: kubevirt.io/hostpath-provisioner
  reclaimPolicy: Delete
  volumeBindingMode: WaitForFirstConsumer
kind: List
metadata:
  resourceVersion: ""
```

PR #469, opened 2023-05-03
Thanos Query: Additional telemetry buckets

jacobbaungard:
This commit adds additional `telemetryDurationQuantiles`, to ensure we have buckets for complicated queries that takes a long time to complete. It uses the default values, as defined upstream in Thanos, and adds 15, 30, 60 and 120. 120 is chosen as the upper bound as that is the timeout set in the Observatorium API.

### DIFF
--- a/resources/services/observatorium-metrics-template.yaml
+++ b/resources/services/observatorium-metrics-template.yaml
@@ -1376,6 +1376,19 @@ objects:
               "service_name": "thanos-query"
             "type": "JAEGER"
           - --query.auto-downsampling
+          - --query.telemetry.request-duration-seconds-quantiles=0.1
+          - --query.telemetry.request-duration-seconds-quantiles=0.25
+          - --query.telemetry.request-duration-seconds-quantiles=0.75
+          - --query.telemetry.request-duration-seconds-quantiles=1.25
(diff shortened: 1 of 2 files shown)


philipgough (approved):
@jacobbaungard looks good to me but I don't see any additional changes via kube-thanos? Did they already go out in #468

jacobbaungard:
> @jacobbaungard looks good to me but I don't see any additional changes via kube-thanos? Did they already go out in #468

Exactly, I rebased after that. Forgot to update the PR description (done now), had just done the commit msg.

saswatamcode (approved):
LGTM! 

So many flags! 🙈 
I wonder if we can modify this in Thanos to just take comma-separated values instead. Might be breaking convention tho

jacobbaungard:
> So many flags! see_no_evil
> I wonder if we can modify this in Thanos to just take comma-separated values instead. Might be breaking convention tho

Yes, it's not great. I already took a brief look at this and couldn't see a way to do it with Kingpin, so we'll probably have to do parsing manually if we'd want to get rid of this pattern (unless I missed something in Kingpin, the docs are not exactly exhaustive)

PR #470, opened 2023-05-03
Triple the resources for Gubernator

douglascamata:
This ensures it can scale up accordingly to properly handle bursts of traffic.


philipgough (approved):
lgtm

PR #480, opened 2023-05-18
Add template for meta-monitoring logging

philipgough:
Adds LokiStack CR to deploy logging backend


resources/services/meta-monitoring/logging-template.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: v1
```
vprashar2929, line 1: Shouldn't this be `template.openshift.io/v1`?

moadz (approved):
LGTM

PR #487, opened 2023-05-25
Add service labels to SLO rules and prune labels

saswatamcode:
Add service label, which is required in App-interface spec, to our SLO rules
Also, prune client, le and container labels

### DIFF
--- a/resources/observability/prometheusrules/pyrra/mst-production-api-metrics-write-availability-slo.yaml
+++ b/resources/observability/prometheusrules/pyrra/mst-production-api-metrics-write-availability-slo.yaml
@@ -9,7 +9,7 @@ metadata:
   creationTimestamp: null
   labels:
     instance: mst-production
-    service: observatorium-api
+    pyrra.dev/service: observatorium-thanos-receive
   name: api-metrics-write-availability-slo
 spec:
   alerting:
--- a/resources/observability/prometheusrules/pyrra/mst-production-api-metrics-write-availability-slo.yaml
+++ b/resources/observability/prometheusrules/pyrra/mst-production-api-metrics-write-availability-slo.yaml
@@ -9,7 +9,7 @@ metadata:
   creationTimestamp: null
   labels:
     instance: mst-production
-    pyrra.dev/service: observatorium-thanos-receive
+    pyrra.dev/service: observatorium-api
   name: api-metrics-write-availability-slo
 spec:
   alerting:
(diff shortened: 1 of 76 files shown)


resources/observability/prometheusrules/pyrra/mst-production-api-metrics-write-latency-slo.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
   creationTimestamp: null
   labels:
     instance: mst-production
-    service: observatorium-api
+    pyrra.dev/service: observatorium-thanos-receive
```
philipgough, line 12: When I look at the expression here, it appears to be using metrics emitted from the obs api so wondering is a change in service name correct here?

resources/observability/prometheusrules/pyrra/mst-production-api-metrics-write-availability-slo.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
   creationTimestamp: null
   labels:
     instance: mst-production
-    service: observatorium-api
+    pyrra.dev/service: observatorium-thanos-receive
```
philipgough, line 12: @saswatamcode should we change the data source in the dashboard urls to use the "externalLabels.cluster" in this PR also?

saswatamcode: This PR only touches SLO alerts, and outputs it for each env separately. Do we need to here then?

resources/observability/prometheusrules/pyrra/mst-production-api-metrics-write-latency-slo.yaml
saswatamcode: Yes, a lot of the metrics are used from obs API, but the alert seems to represent the /receive endpoint. So I wanted to represent the name of the exact microservice here, instead of API, so that it is indicative of what to check exactly. 

If we change to just obs-api due to the metric, then many alerts around query and rules would also need to be service: obs-api.

Wdyt?

philipgough: Yeah, makes sense

resources/observability/prometheusrules/pyrra/mst-production-api-metrics-write-availability-slo.yaml
philipgough: I guess not

philipgough (approved):

PR #488, opened 2023-05-25
add thanos receive limits configmap

thibaultmg:
Adds an openshift template for generating thanos receive limits configmap.

Couldn't do something fancier than that. Was unable to embed yaml in yaml. The whole configuration must be embedded into the RECEIVE_LIMITS parameter. 
But does the job.

### DIFF
--- a/resources/services/observatorium-metrics-limits-template.yaml
+++ b/resources/services/observatorium-metrics-limits-template.yaml
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
   annotations:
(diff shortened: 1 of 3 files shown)


saswatamcode (approved):
LGTM!

resources/services/observatorium-metrics-limits-template.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
```
vprashar2929, line 1: Shouldn't `apiVersion: v1` be `apiVersion: template.openshift.io/v1` ?

thibaultmg: good point. fixed

PR #492, opened 2023-05-27
Add support to deploy meta monitoring via launch script

vprashar2929:
This PR:
* Add support to deploy Loki stack via launch script on any OCP cluster.
* Add support to deploy prerequisites like OpenShift logging and Loki operator required for meta-monitoring template.
* Add support to deploy `clusterlogging` and `clusterlogforwarder` as part of meta monitoring stack.

### DIFF
--- a/services/minio-template.jsonnet
+++ b/services/minio-template.jsonnet
@@ -15,6 +15,7 @@ local minio = (import 'github.com/observatorium/observatorium/configuration/comp
         spec+: {
           containers: [
             super.containers[0] {
+              image: '${IMAGE}:${IMAGE_TAG}',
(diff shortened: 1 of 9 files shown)


vprashar2929:
Yes, this was also my concern but this commit doesn't touch CI. It's just deploying the stack via our [launch script ](https://github.com/rhobs/configuration/blob/main/tests/launch.sh)that we run on a cluster-bot cluster. For CI I am planning to add something like `oc process -f logging-template.yaml` which will not deploy the template but validate syntactically against the running cluster. This will avoid deploying operator's and other dependencies on CI.

PR #496, opened 2023-05-31
Add rule-query to rules & dashboards

saswatamcode:
Add split ruler query to our rules/dashboards for observability


philipgough (approved):
lgtm

PR #500, opened 2023-06-02
receive: fix receive limits configmap generation

thibaultmg:
Using double curly braces fails on app-sre side. It expects a json as parameter and fails otherwise.
Changing to single curly braces. As per my tests, openshift template correctly generates multiline string in resulting configmap for the RECEIVE_LIMITS parameter.


coleenquadros:
lgtm

saswatamcode (approved):

PR #507, opened 2023-06-07
Fix makefile

vprashar2929:
When running `make validate` on macOS the command gets failed as the [latest](https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest/) OC version artifacts don't contain tar's with `darwin`.


jacobbaungard (approved):
Looks good :+1:

philipgough (approved):

PR #508, opened 2023-06-08
Fixing neverFiringAlerts to return false expression

moadz:
Currently the `NeverFiring` alerts always fire as `vector(0)` returns a single sample (of value `0`), this fixes that.


saswatamcode (approved):
LGTM!

PR #512, opened 2023-06-14
RHOBS-781: Reducing error budget SLO alerts to 95% from 99.5%

moadz:
The change to introduce these burn alerts at 99.5% sent a lot of alerts to app-sre and thus was reverted. As per https://docs.google.com/document/d/1fogL6R1eN4dwf_0Bz0K4S4okwJzJIX1uqyTi9ZCsvWw/edit we're going to investigate the likelihood of it firing at a lower threshold and work our way up.

### DIFF
--- a/configuration/observatorium/slo.go
+++ b/configuration/observatorium/slo.go
@@ -13,11 +13,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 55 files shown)


saswatamcode (approved):
LGTM!

PR #513, opened 2023-06-14
Fix csv status check

vprashar2929:
This PR fixes the status check of `csv` inside the launch script.


douglascamata (approved):
LGTM :shipit:

PR #514, opened 2023-06-14
Refactor Observatorium Api dashoard

coleenquadros:
Observatorium API is deployed as 
`observatorium-observatorium-api` - api is in observatorium-production/stage namespace and thanos components in observatorium-metrics-production/stage namespace
`observatorium-observatorium-mst-api` -Both  api and Thanos components deployed in observatorium-mst-production/stage namespace

Changing the drop down selector to pick up the api job and namespace


coleenquadros:
test dashoard - https://grafana.stage.devshift.net/d/7hXbx6_4k/api2?orgId=1&refresh=1m&var-datasource=telemeter-prod-01-prometheus&var-job=observatorium-observatorium-mst-api&var-handler=All

douglascamata:
General comment unrelated to the PR, but good to have it documented: I think we should move these dashboards to somewhere inside the https://github.com/observatorium org.

coleenquadros:
added namespace selector - https://grafana.stage.devshift.net/d/3yjYCel4z/api4?orgId=1&refresh=1m&var-datasource=telemeter-prod-01-prometheus&var-namespace=observatorium-mst-production&var-job=observatorium-observatorium-mst-api&var-handler=All

PR #518, opened 2023-06-15
Removing the parca scrape targets for Loki logs 

moadz:
Logs is no longer deployed by default in our environments. This config causes Parca to crashloop.


saswatamcode (approved):
LGTM!

PR #521, opened 2023-06-16
RHOBS-781: Removing _code_ filtering on sum expression to fix incorrect SLO calculation

moadz:
### Why? 
* When we released these SLOs into production, we noticed they were very sensitive 
* Upon investigating the underlying query we realised the Pyrra was incorrectly generating promql for the underlying recording rules that determine the error rate

### Impact of this change
* We lose per-error code aggregations on our alerts, so the alert will no longer be
* An argument can be made that the total error budget alert shouldn't be aggregated by code anyway, because the burnrate will be skewed, as you could have a cumulative error rate that exceeds your SLO, but not one individual code doing so
 
 
# Context
### What is the burn rate calculation? 
For a time range `t` `burnrate = rate_of_errors[t] / rate_of_requests[t]`

So, for example for `receive` burn rate over `6h` the following expression represents the: 
* Rate of errors `http_requests_total{code=~"^5..$",group="metricsv1",handler="receive",job="observatorium-observatorium-mst-api"}` 
* Rate of requests `http_requests_total{code!~"^4..$",group="metricsv1",handler="receive",job="observatorium-observatorium-mst-api"}`

Because the denominator includes a `code` label expression that != the numerator, pyrra assumes a `sum by (code)` is necessary. This is flawed, as your denominator should always be a single vector, not a matrix. The result of this logic is the following: 

```
sum by (code) (rate(http_requests_total{code=~"^5..$",group="metricsv1",handler="receive",job="observatorium-observatorium-mst-api"}[6h]))
\
sum by (code) (rate(http_requests_total{code!~"^4..$",group="metricsv1",handler="receive",job="observatorium-observatorium-mst-api"}[6h]))
```
**Which will always return `1` for any non-zero error value.** An error rate of `1` means 100% of your requests are erroring (which is not correct)

Ideally the expression should be the following to preserve the error code: 
```
(sum by (code) (rate(http_requests_total{code=~"^5..$",group="metricsv1",handler="receive",job="observatorium-observatorium-mst-api"}[6h])))
/ on() group_left() sum(rate(http_requests_total{code!~"^4..$",group="metricsv1",handler="receive",job="observatorium-observatorium-mst-api"}[6h]))
```

### Why do we exclude `4xx` status codes from our total requests in the first place? And what is the impact of including them? 
We don't want to skew the error rate based on genuine client errors (4xx). So the impact of excluding them will be that the error rate is going to a function of `5xx` / `2|4|5xx` instead of `2|5xx` which may reduce our error rate further. 

E.g. Suppose we have a 99% SLO on availability for a given time period, and we have 99 `2xx` requests and 1 `5xx` requests  
* We have a 2% error rate (1/(99 + 1))
* SLO is about to be breached (99% success avail) 

Suppose a tenant were to suddenly spam our requests with 100 `404` due to hitting the incorrect `tenant` URL in their request: 
* We would have a 0.5% error rate (1/(99 + 100 + 1))
* SLO is not close to being breached (99.5% success avail)

we don't want that traffic to reduce the proportion of error relative to it. 

An argument can be made for either case, including `4xx` (as they are legitimate requests and thus the proportion of errors is still functionally correct) and excluding them (we want a purer view of successful traffic vs internal errors w/ out consideration for other client issues).


saswatamcode (approved):
Thanks for awesome explanation and effort!
LGTM!

PR #528, opened 2023-06-22
initial rhobs utilization dashboard with query front end row

coleenquadros:
https://issues.redhat.com/browse/RHOBS-704


coleenquadros:
staging dashboard - https://grafana.stage.devshift.net/d/G1nb8JX4z/rhobs-instance-utilization-overview?orgId=1&refresh=10s&var-datasource=telemeter-prod-01-prometheus&var-namespace=observatorium-metrics-production&var-interval=5m

Ive given the grid positions for the panels such a way that there are 4 panels per row, but when I copy paste the generated JSON the panels gets stacked on one line.

observability/dashboards/rhobs-instance-utilization-overview.libsonnet
```diff
@@ -0,0 +1,127 @@
...
+              'cpu usage system {{instance}}',
+            ]
+          ) +
+          { yaxes: g.yaxes('percent'), gridPos: { x: 6, y: 7, w: 6, h: 6 } },
+        )
+        .addPanel(
+          g.panel('Pod/Container Restarts') +
+          g.queryPanel(
+            [
+              'kube_pod_container_status_restarts_total{namespace="$namespace", container=\'thanos-query-frontend\'}',
```
philipgough, line 105: i wonder should we calculate rate for this counter?

observability/dashboards/rhobs-instance-utilization-overview.libsonnet
```diff
@@ -0,0 +1,127 @@
...
+            ],
+            [
+              'alloc all {{instance}}',
+              'alloc heap {{instance}}',
+              'alloc rate all {{instance}}',
+              'alloc rate heap {{instance}}',
+              'inuse heap {{instance}}',
+              'inuse stack {{instance}}',
+            ]
+          ) +
```
douglascamata, line 86: For this panel could we simplify things a bit and graph only the working set, for example? There's a good metric for this: `container_memory_working_set_bytes`.

We very rarely look at any Go specific memory metric and when we do mostly we care about in-use heap. So they can stay in the specific dashboard. 👍

observability/dashboards/rhobs-instance-utilization-overview.libsonnet
```diff
@@ -0,0 +1,127 @@
...
+    title: error 'must provide title for Thanos Query Frontend dashboard',
+    dashboard:: {
+      selector: std.join(', ', config.dashboard.selector + ['job=~"observatorium-thanos-query-frontend"']),
+      dimensions: std.join(', ', config.dashboard.dimensions + ['job']),
+      pod: 'observatorium-thanos-query-frontend.*',
+    },
+  },
+
+  local queryFrontendHandlerSelector = utils.joinLabels([thanos.queryFrontend.dashboard.selector, 'handler="query-frontend"']),
+  local queryFrontendOpSelector = utils.joinLabels([thanos.queryFrontend.dashboard.selector, 'op="query_range"']),
```
douglascamata, line 41: What about instant queries that the query frontend might handle? 🤔

douglascamata (changes requested):
This is missing the link to the query frontend dashboard, no? Do you plan to add it in a follow up PR?

observability/dashboards/rhobs-instance-utilization-overview.libsonnet
coleenquadros: I picked this up from the mixin itself. So if one were to move to the original component dashboard then this panel will end up being different.

douglascamata: Yes, it will be different there. But was there, with all these metrics, is not useful. The chart looks too crowded. Also this chart has only these metrics in Thanos because there it was decided to only use metrics exported from Thanos itself. As `container_memory_working_set_bytes` comes from a different place, it wasn't included.

observability/dashboards/rhobs-instance-utilization-overview.libsonnet
coleenquadros: Removed the `query_range` filter.. Will need to update upstream mixin

observability/dashboards/rhobs-instance-utilization-overview.libsonnet
```diff
@@ -0,0 +1,124 @@
...
+          g.panel('Duration', 'Shows how long has it taken to handle requests in quantiles.') +
+          g.latencyPanel('http_request_duration_seconds', queryFrontendHandlerSelector, thanos.queryFrontend.dashboard.dimensions) +
+          g.addDashboardLink(thanos.queryFrontend.dashboard.title) +
+          { gridPos: { x: 18, y: 1, w: 6, h: 6 } },
+        )
+        .addPanel(
+          g.panel('Memory Used') +
+          g.queryPanel(
+            [
+              '(container_memory_working_set_bytes{container="observatorium-thanos-query-frontend", namespace="$namespace"}[$interval]) / (1024 * 1024)',
```
douglascamata, line 75: This is not a range vector, so the `[$interval]` part won't work and can be safely removed.

douglascamata (changes requested):
I imported the changed dashboard into our staging environment and noticed a problem with the memory used widget. 

The layout also looks off. Seems like charts are crammed into a very small space. Can you please give them more space so that they look better? It's okay to have more than 1 row/line of widgets.

coleenquadros:
> I imported the changed dashboard into our staging environment and noticed a problem with the memory used widget.
> 
> The layout also looks off. Seems like charts are crammed into a very small space. Can you please give them more space so that they look better? It's okay to have more than 1 row/line of widgets.

Thats where I provided the grid positions in the panels to allow the panels to be split as 4 panels in one line. But when I copy the JSON it gets moved to one line

observability/dashboards/rhobs-instance-utilization-overview.libsonnet
```diff
@@ -0,0 +1,125 @@
...
+          { yaxes: g.yaxes('percent'), gridPos: { x: 6, y: 7, w: 6, h: 6 } },
+        )
+        .addPanel(
+          g.panel('Pod/Container Restarts') +
+          g.queryPanel(
+            [
+              'increase(kube_pod_container_status_restarts_total{namespace="$namespace", container=\'thanos-query-frontend\'}[$interval])',
+            ],
+            [
+              'pod {{instance}}',
```
douglascamata, line 105: It doesn't seem like this metric has an `instance` label. We have to use the `pod` label here, I believe. The same applies to the cpu metric just above.

Note that this is already done in the memory metric.

douglascamata (changes requested):
One final set of changes and I think we can merge this. It's looking pretty good! :D

observability/dashboards/rhobs-instance-utilization-overview.libsonnet
```diff
@@ -0,0 +1,125 @@
...
+    },
+  },
+
+  local queryFrontendHandlerSelector = utils.joinLabels([thanos.queryFrontend.dashboard.selector, 'handler="query-frontend"']),
+  local queryFrontendOpSelector = utils.joinLabels([thanos.queryFrontend.dashboard.selector, 'op=~".*"']),
+  dashboard:: {
+    data:
+      g.dashboard('RHOBS Instance Utilization Overview')
+      .addRow(
+        g.row('Query Frontend Overview')
```
douglascamata, line 47: We are missing the network throughput charts here, as mentioned in the JIRA ticket.

observability/dashboards/rhobs-instance-utilization-overview.libsonnet
```diff
@@ -0,0 +1,125 @@
...
+          g.httpQpsPanel('http_requests_total', queryFrontendHandlerSelector, thanos.queryFrontend.dashboard.dimensions) +
+          g.addDashboardLink(thanos.queryFrontend.dashboard.title) +
+          { gridPos: { x: 0, y: 1, w: 6, h: 6 } },
+        )
+        .addPanel(
+          g.panel('Rate of queries', 'Shows rate of queries passing through Query Frontend') +
+          g.httpQpsPanel('thanos_query_frontend_queries_total', queryFrontendOpSelector, thanos.queryFrontend.dashboard.dimensions + ',op') +
+          g.addDashboardLink(thanos.queryFrontend.dashboard.title) +
+          { gridPos: { x: 6, y: 1, w: 6, h: 6 } },
+        )
```
douglascamata, line 59: Aren't these two the same thing?

coleenquadros: Yes, they are but represented different in the panels, with one panel using the handler="query-frontend" and the other panel using "op=query_range". We can choose to keep one 

This is taken from the existing thanos query front end dashboard - https://grafana.stage.devshift.net/d/7c68ed2ef2355474f058dd27f0471f7a/thanos-query-frontend?orgId=1&refresh=10s

douglascamata: Let's pick one then. If upstream there wasn't the op filter, these charts would mostly match 100%.

douglascamata (approved):
Nice one, LGTM :shipit:

PR #532, opened 2023-06-26
Reorganize tests folder

vprashar2929:
This PR reorganize the tests folder since we added support for deploying more components on any OCP cluster as well as template validations on CI Hence reducing the clutter.


douglascamata (approved):
LGTM :shipit:

PR #534, opened 2023-07-03
Add row in instance utilization dashboard for Thanos Receive

douglascamata:
I'm planning to follow up with a PR allowing the `job` to be a dashboard variable and used in all the charts. This should help with filtering information.

### DIFF
--- a/observability/dashboards/rhobs-instance-utilization-overview.libsonnet
+++ b/observability/dashboards/rhobs-instance-utilization-overview.libsonnet
@@ -144,7 +144,7 @@ function() {
           g.panel('Network Traffic') +
           g.queryPanel(
             [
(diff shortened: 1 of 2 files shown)


coleenquadros:
LGTM

I have created functions for the common resource utlization panels(memory,n/w,cpu,restarts) so that they are not repeated. If you'd like you can add them to your diff and I can update my branch later

https://github.com/rhobs/configuration/pull/535/files

douglascamata:
@coleenquadros hummm, I see. I predicted we would be better off with helpers for some charts that will be repeated. But introducing them before we finish the creation of all the rows would case conflicts that can be potentially difficult to solve. 

I propose that we finish the implementation of the rows and then start refactoring. What do you think, @coleenquadros?

coleenquadros (approved):
Lgtm

saswatamcode (approved):
LGTM!

coleenquadros (approved):

PR #542, opened 2023-07-06
Alertmanager and Ruler query overview

coleenquadros:
[ttps://issues.redhat.com/browse/RHOBS-705](https://issues.redhat.com/browse/RHOBS-705)
https://issues.redhat.com/browse/RHOBS-753

test dashboard - https://grafana.stage.devshift.net/d/T0nMNU9Vz/rhobs-instance-utilization-overview-test-10?var-datasource=telemeter-prod-01-prometheus&var-namespace=observatorium-metrics-production&var-job=All&var-interval=5m

PR #543, opened 2023-07-07
Fix layout issue in instance dashboard

vprashar2929:
In thanos mixin [builder.libsonnet](https://github.com/thanos-io/thanos/blob/main/mixin/lib/thanos-grafana-builder/builder.libsonnet) we import [grpc.libsonnet](https://github.com/thanos-io/thanos/blob/main/mixin/lib/thanos-grafana-builder/grpc.libsonnet), [http.libsonnet](https://github.com/thanos-io/thanos/blob/main/mixin/lib/thanos-grafana-builder/http.libsonnet), slo.libsonnet and uses their functions like `httpQpsPanel`, `httpErrPanel`, `latencyPanel` etc for creating respective panels in our libsonnet. Since these function's don't have the property to change the span size and without that, they pick up the default values from [this](https://github.com/grafana/jsonnet-libs/blob/f33dff93db677a32303630c3e0910cf6d46a92cc/grafana-builder/grafana.libsonnet#L182). It's best to remove the span property completely as it interferes with the default layout sizing which grafana provides as removing the span property allows grafana to dynamically size the panels inside a row.

[Link](https://grafana.stage.devshift.net/d/ltvkT_r4z/vprashar-test?orgId=1&refresh=10s&var-datasource=rhobsp02ue1-prometheus&var-namespace=observatorium-mst-production&var-job=All&var-interval=5m&from=1688715625166&to=1688719225166) to dashboard using generated JSON

### DIFF
--- a/observability/dashboards/rhobs-instance-utilization-overview.libsonnet
+++ b/observability/dashboards/rhobs-instance-utilization-overview.libsonnet
@@ -109,39 +109,39 @@ function() {
       .addRow(
         g.row('Receive Overview')
         .addPanel(
-          g.panel('Rate of requests', 'Shows rate of requests against Receive for the given time') +
-          g.httpQpsPanel('http_requests_total', receiveHandlerSelector, thanos.receive.dashboard.dimensions) +
+          g.panel('Rate of requests', 'Shows rate of requests against Receive for the given time') { span:: 0 } +
+          g.httpQpsPanel('http_requests_total', receiveHandlerSelector, thanos.receive.dashboard.dimensions) { span:: 0 } +
           g.addDashboardLink(thanos.receive.dashboard.title) +
           g.stack
         )
         .addPanel(
-          g.panel('Errors', 'Shows ratio of errors compared to the total number of handled requests against Receive.') +
-          g.httpErrPanel('http_requests_total', receiveHandlerSelector, thanos.receive.dashboard.dimensions) +
+          g.panel('Errors', 'Shows ratio of errors compared to the total number of handled requests against Receive.') { span:: 0 } +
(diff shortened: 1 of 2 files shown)


coleenquadros (approved):

coleenquadros:
I think two more rows need to be updated[(https://github.com/rhobs/configuration/pull/542)](https://github.com/rhobs/configuration/commit/ec28b919c84cbc116d970dcddd05c95fd4ab1d9b)

thibaultmg (approved):
Awesome! 
Maybe we should fix the faulty panels upstream? It would still be nice to be able to define our own layout...

vprashar2929:
Updated [Dashboard](https://grafana.stage.devshift.net/d/HK4ALl9Vk/vprashar-test?orgId=1&refresh=10s&var-datasource=rhobsp02ue1-prometheus&var-namespace=observatorium-mst-production&var-job=All&var-interval=5m&from=1688713941976&to=1688724741976)

coleenquadros (approved):

vprashar2929:
> Maybe we should fix the faulty panels upstream? It would still be nice to be able to define our own layout...

Initially, I also thought to fix this upstream but if everyone else is using why didn't they face the similar layout problem which we did? Is it because we use a different Grafana version or is it something related to our environment? I am still not clear on this that's why I didn't touch upstream. But yes if this is a genuine case then we should update upstream as well

saswatamcode (approved):
LGTM! 

I don't quite get how others upstream aren't affected by this. Maybe they fix it after import as well, or just copy over dashboard yaml and fix? In any case, would be great to have such a fix templated upstream

PR #549, opened 2023-07-12
refactoring rhobs instance overview dashboard

coleenquadros:
- Resource utilization panels are now functions
- added `collapse = true `attribute to all rows 
https://issues.redhat.com/browse/RHOBS-812

test dashboard - https://grafana.stage.devshift.net/d/nSVblSCVz/rhobs-instance-utilization-overview-test-3?orgId=1&refresh=10s

### DIFF
--- a/observability/dashboards/rhobs-instance-utilization-overview.libsonnet
+++ b/observability/dashboards/rhobs-instance-utilization-overview.libsonnet
@@ -42,24 +42,24 @@ function() {
     ),
 
   queryFrontend:: {
-    selector: error 'must provide selector for Thanos Query Frontend dashboard',
-    title: error 'must provide title for Thanos Query Frontend dashboard',
     dashboard:: {
       title: config.queryFrontend.title,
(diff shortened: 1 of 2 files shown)


observability/dashboards/rhobs-instance-utilization-overview.libsonnet
```diff
@@ -79,39 +77,103 @@ function() {
...
+      [
+        '(container_memory_working_set_bytes{container="%s", pod=~"%s", namespace="$namespace"})' % [
+          container,
+          pod,
+        ],
+      ],
+      [
+        'memory usage system {{pod}}',
+      ]
+    ) { span:: 0 },
```
douglascamata, line 134: I think we don't need the `span` in both places, as in the end everything's merged together. This applies to all the panels we have.

douglascamata (changes requested):
Awesome refactor! It'll be very easy for us to change and further improve these from now on. Great job. 💪 

There's just one thing that I believe we should improve on.

douglascamata (approved):
LGTM :shipit:

PR #551, opened 2023-07-17
MON-3274: sync telemeter config with apiserver_list_watch_request_success_total:rate:sum

p0lyn0mial:
I just ran `make whitelisted_metrics && make` as described in Following https://rhobs-handbook.netlify.app/products/openshiftmonitoring/telemetry.md

This PR adds `apiserver_list_watch_request_success_total:rate:sum` to the telemeter configuration. 
The metrics was added in https://github.com/openshift/cluster-monitoring-operator/pull/2044

Update: to pass `git diff --exit-code` I had to run `make manifests` (this could be reflected in the docs)


jan--f:
This also adds another telemetry metrics added in https://github.com/openshift/cluster-monitoring-operator/pull/2030
cc @rexagod 
Just leaving this here for future us.

jan--f (approved):

PR #553, opened 2023-07-17
Update SLOs for split querier

thibaultmg:
SLOs are updated to take into account the current deployment topology that splits the queriers for ad hoc queries and rules evaluation.

- availability and latencies SLOs are added for the rule-query 
- new observatorium-up deployment added to generate latency metrics
-> I could not reuse the current deployment as it targets the query frontend

### DIFF
--- a/configuration/observatorium/queries-ruler.libsonnet
+++ b/configuration/observatorium/queries-ruler.libsonnet
@@ -1,0 +1,16 @@
+{
+  queries: [
+    {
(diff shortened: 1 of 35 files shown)


douglascamata (approved):
LGTM :shipit:

saswatamcode (approved):
Thanks this looks good! Awesome work!

I think this would be painful to rollout tho, so @thibaultmg it might make sense to note down what the steps for deploying every part of this are. 🙂

PR #555, opened 2023-07-18
Remove all namespace/job  option on rhobs overview dashboard

coleenquadros:
https://grafana.stage.devshift.net/d/zuUu4sCVz/rhobs-instance-utilization-overview-test-3?orgId=1&refresh=10s


coleenquadros:
this means that certain panels will load all the jobs for all the namespace because the query doesnt select a specific job and the panel is not going to make sense

observability/dashboards/rhobs-instance-utilization-overview.libsonnet
```diff
@@ -45,7 +44,7 @@ function() {
   queryFrontend:: {
     dashboard:: {
       title: config.queryFrontend.title,
-      selector: std.join(', ', config.dashboard.selector + ['job=~"$job"']),
```
douglascamata, line 48: @coleenquadros do you think we can fix the source of `config.dashboard.selector` instead of modifying it here?

douglascamata: I see. They do this there because in their mixin dashboards they aren't showing the "All" option for the namespace selector. 🤔

Do you think it's valuable to keep the "All" option or should we default to match the thanos-mixin dashboards and remove this option?

coleenquadros: It all depends on how a support engineer would look at it.. he should be aware that he has to select a namespace and a job, or if we give the all option on the namespace then each of the component rows should display the right job because all option on the job right now is going to be a mess

douglascamata: Agreed. I think seeing all namespaces will be super noisy in general, unless the deployment is really small. I'm inclined towards matching the behavior of the upstream dashboards that we already use to keep the consistency.

coleenquadros: so we remove the all option in namespace then?

douglascamata: Yup, I think it's okay to do it.

coleenquadros:
@douglascamata changed the PR to remove all option on namespace and job

douglascamata:
@coleenquadros the job one is fine to keep, as the matcher it uses is the regex one (`=~`).

douglascamata (approved):
LGTM :shipit:

PR #563, opened 2023-07-26
Add gubernator dashboard

thibaultmg:
Adds new Gubernator dashboard.
Tried to simplify the jsonnet file. 
Only the process_cpu metric is not working. I do not know why it isn't generated.

Here is the dashboard: https://grafana.stage.devshift.net/d/q24ceK3Vz/observatorium-gubernator?orgId=1&refresh=10s&var-datasource=app-sre-stage-01-prometheus&var-namespace=observatorium-mst-stage&var-interval=5m


coleenquadros:
Is there a test dashboard that we can view?

thibaultmg:
> Is there a test dashboard that we can view?

Yes, updated the description 😉

douglascamata (approved):
LGTM

PR #572, opened 2023-08-02
Operations: Add tooling/docs/monitoring for thanos bucket inspect/replicate

philipgough:
https://issues.redhat.com/browse/RHOBS-816

Screenshot 2023-08-02 at 11 56 20

### DIFF
--- a/resources/operations/bucket-inspect/job-template.yaml
+++ b/resources/operations/bucket-inspect/job-template.yaml
@@ -1,0 +1,57 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: rhobs-thanos-bucket-inspect
+  labels:
+    app.kubernetes.io/name: thanos-bucket-inspect
+    app.kubernetes.io/part-of: observatorium
+description: |
+  Inspect data in an object storage provider bucket
+parameters:
+  - name: NAME
+    description: The name of the Job.
+    value: 'thanos-bucket-inspect'
+  - name: NAMESPACE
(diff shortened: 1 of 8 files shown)


resources/operations/bucket-inspect/job-template.yaml
saswatamcode, line 1: This is the one-off inspect?

resources/operations/bucket-replicate/README.md
```diff
@@ -0,0 +1,50 @@
+# What
+
+This template deploys [Thanos Bucket Inspect](https://thanos.io/tip/components/tools.md/#bucket-insepct)
+as a Kubernetes Job.
+
+# SOP
+
+> **_NOTE:_**  Before running this Job, if you wish to track progress via logs,
```
saswatamcode, line 8: I'd also add a note here, that depending on number of blocks in the `from` bucket...this job might be over even before single Prometheus scrape.

saswatamcode (approved):
Thanks! Generally LGTM!

resources/operations/bucket-inspect/job-template.yaml
philipgough: yes one of inspect. would suggest we run it on the original source and then compare iteratively against the cronjob. Also very useful for testing your s3 config and secrets

resources/operations/bucket-replicate/README.md
philipgough: yeah, good idea

PR #582, opened 2023-08-27
K8s job for using rclone to copy buckets instead of thanos-replicate

moadz:
cc: @philipgough


saswatamcode (approved):
LGTM! What limitations did you notice with `replicate`?

resources/operations/rclone-bucket-replicate/job-template.yaml
```diff
@@ -0,0 +1,74 @@
...
+description: |
+  Replicate data between object storage provider buckets
+parameters:
+  - name: NAME
+    description: The name of the Job.
+    value: 'rclone-bucket-replicate'
+  - name: NAMESPACE
+    description: The namespace where the Job should run.
+    value: 'observatorium-operations'
+  - name: IMAGE_TAG
```
philipgough, line 17: this still points at latest Thanos so we should fix that and pass it in down below instead of `latest`

moadz: Fixed

moadz:
> @saswatamcode did you notice with replicate? Usually not an issue for block storage syncing, but for en-masse migration it's quite slow. 

We also don't support parallel uploaders (which would be a good feature if threadsafe): 
This: https://github.com/observatorium/thanos-replicate/issues/22

philipgough (approved):

PR #583, opened 2023-08-30
Fix missing metrics from telemeter memcached

thibaultmg:
Adds a namespace selector to the memcached serviceMonitor


philipgough (approved):

moadz (approved):
LGTM

PR #587, opened 2023-08-31
Update obsctl-reloader to be picked up by Prometheus

douglascamata:
It was missing the `prometheus: app-sre` label.


coleenquadros:
lgtm

PR #602, opened 2023-09-21
WIP: Create go configuration rhobs

thibaultmg:


mimic.go
```diff
@@ -2,15 +2,18 @@ package main
...
 
 func main() {
 	gen := mimic.New()
 
 	defer gen.Generate()
 
-	cfgobservatorium.GenSLO(gen.With("observability", "prometheusrules", "pyrra"), gen.With("observability", "prometheusrules"))
+	// cfgobservatorium.GenSLO(gen.With("observability", "prometheusrules", "pyrra"), gen.With("observability", "prometheusrules"))
+
+	// cfgobservatorium.GenerateRBAC(gen.With(".tmp", "tenants"))
```
saswatamcode, line 15: Why disable them?

thibaultmg: Does not work anymore locally, I have to check what is going on 🤷‍♂️ . This is a draft PR, just to show the direction I am going. See if there are some early comments while we wait for other components to be added.

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
```diff
@@ -0,0 +1,298 @@
...
+        accessModes:
+        - ReadWriteOnce
+        resources:
+          requests:
+            storage: 500Gi
+        storageClassName: gp2
+      status: {}
+  status:
+    availableReplicas: 0
+    replicas: 0
```
douglascamata, line 298: probably don't need this?

services_go/observatorium/observatorium.go
```diff
@@ -0,0 +1,96 @@
...
+type InstanceConfiguration struct {
+	Cluster             string
+	Instance            string
+	Namespace           string
+	Tenants             []TenantInstanceConfiguration
+	ThanosStoreReplicas int32
+}
+
+// PostProcessFunc is a function that can be applied to a Kubernetes object after it has been generated.
+type PostProcessFunc func(obj runtime.Object)
```
douglascamata, line 43: What's the purpose of this post-processing pattern here? It seems clunky and weird to me. 

For example, I see that `NewObservatorium` is collecting the post-process functions of all components, then it generates the manifests and runs the functions to update them. But why can't these modifications be applied directly to the functions generating each components (`makeStore`, `makeCompactor`)?

douglascamata (commented):
I think you need to update this to reflect the fact that we're running Tempo now instead of Jaeger, right @philipgough?

Also left some questions.

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,279 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  creationTimestamp: null
+  name: observatorium-metrics-compact
+objects:
+- apiVersion: v1
+  data:
+    session_secret: c2VjcmV0
```
philipgough, line 9: whats the purpose of this?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+- apiVersion: monitoring.coreos.com/v1
+  kind: ServiceMonitor
+  metadata:
+    creationTimestamp: null
+    labels:
+      app.kubernetes.io/component: database-compactor
+      app.kubernetes.io/instance: observatorium
+      app.kubernetes.io/name: thanos-compact
+      app.kubernetes.io/part-of: observatorium
+      app.kubernetes.io/version: v0.32.3
```
philipgough, line 73: this will need an `prometheus: 'app-sre'` label in order to be discovered

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+    name: observatorium-thanos-compact
+    namespace: openshift-monitoring
```
philipgough, line 75: Is this correct? I think we would need to land this in openshift-customer-monitoring?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+      matchNames:
+      - rhobs
+    selector:
+      matchLabels:
+        app.kubernetes.io/component: database-compactor
+        app.kubernetes.io/instance: observatorium
+        app.kubernetes.io/name: thanos-compact
+        app.kubernetes.io/part-of: observatorium
+- apiVersion: apps/v1
+  kind: StatefulSet
```
philipgough, line 96: I think we will need a PodDisruptionBudget for this?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+                  - key: app.kubernetes.io/name
+                    operator: In
+                    values:
+                    - thanos-compact
+                  - key: app.kubernetes.io/instance
+                    operator: In
+                    values:
+                    - observatorium
+                namespaces:
+                - observatorium
```
philipgough, line 142: is this correct? should be rhobs I think?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+                topologyKey: kubernetes.io/hostname
+              weight: 100
+        containers:
+        - args:
+          - compact
+          - --compact.concurrency=1
```
philipgough, line 148: perhaps this should be a parameter with a default of 1?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+          - --data-dir=/var/thanos/compactor
+          - --deduplication.replica-label=replica
+          - --delete-delay=24h0m0s
+          - --downsample.concurrency=1
+          - --log.format=logfmt
+          - --log.level=warn
```
philipgough, line 154: maybe a parameter also?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+            "namespace": "rhobs"}'
+          - '-openshift-delegate-urls={"/": {"resource": "namespaces", "verb": "get",
+            "name": "rhobs", "namespace": "rhobs"}}'
+          - -tls-cert=/etc/tls/private/tls.crt
+          - -tls-key=/etc/tls/private/tls.key
+          - -client-secret-file=/var/run/secrets/kubernetes.io/serviceaccount/token
+          - -cookie-secret-file=/etc/proxy/secrets/session_secret
+          - -openshift-ca=/etc/pki/tls/cert.pem
+          - -openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+          image: quay.io/openshift/origin-oauth-proxy:v4.8.0
```
philipgough, line 223: just fyi 4.13.0 is the latest release

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+          - name: AWS_SECRET_ACCESS_KEY
+            valueFrom:
+              secretKeyRef:
+                key: aws_secret_access_key
+                name: rhobs-thanos-s3
+          - name: OBJSTORE_CONFIG
+            valueFrom:
+              secretKeyRef:
+                key: thanos.yaml
+                name: rhobs-thanos-objectstorage
```
philipgough, line 176: where are these secrets coming from? should they not be part of the template itself?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
```diff
@@ -0,0 +1,298 @@
...
+  metadata:
+    creationTimestamp: null
+    labels:
+      app.kubernetes.io/component: object-store-gateway
+      app.kubernetes.io/instance: observatorium
+      app.kubernetes.io/name: thanos-store
+      app.kubernetes.io/part-of: observatorium
+      app.kubernetes.io/version: v0.32.3
+    name: observatorium-thanos-store
+    namespace: openshift-monitoring
```
philipgough, line 76: same comment with respect to the missing label and namespace issue

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
```diff
@@ -0,0 +1,298 @@
...
+      matchNames:
+      - rhobs
+    selector:
+      matchLabels:
+        app.kubernetes.io/component: object-store-gateway
+        app.kubernetes.io/instance: observatorium
+        app.kubernetes.io/name: thanos-store
+        app.kubernetes.io/part-of: observatorium
+- apiVersion: apps/v1
+  kind: StatefulSet
```
philipgough, line 97: also will need a PDB

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
```diff
@@ -0,0 +1,298 @@
...
+                  - key: app.kubernetes.io/name
+                    operator: In
+                    values:
+                    - thanos-store
+                  - key: app.kubernetes.io/instance
+                    operator: In
+                    values:
+                    - observatorium
+                namespaces:
+                - observatorium
```
philipgough, line 143: same comment as before

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
```diff
@@ -0,0 +1,298 @@
...
+          - name: AWS_SECRET_ACCESS_KEY
+            valueFrom:
+              secretKeyRef:
+                key: aws_secret_access_key
+                name: rhobs-thanos-s3
+          - name: OBJSTORE_CONFIG
+            valueFrom:
+              secretKeyRef:
+                key: thanos.yaml
+                name: rhobs-thanos-objectstorage
```
philipgough, line 172: same comment wrt to the secrets

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
```diff
@@ -0,0 +1,298 @@
...
+              cpu: "4"
+              memory: 20Gi
+          terminationMessagePolicy: FallbackToLogsOnError
+          volumeMounts:
+          - mountPath: /var/thanos/store
+            name: data
+          - mountPath: /etc/config
+            name: hashmod-config
+        - args:
+          - --reporter.grpc.host-port=dns:///jaeger-collector-headless.observatorium-tools.svc:14250
```
philipgough, line 206: please see https://github.com/rhobs/configuration/commit/225b49524262b74f899941ec6dd43e861f2fa979 for a change here

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
philipgough: PodDisruptionBudget, mentioned above too by name in full so shortened here but sorry for the confusion

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
thibaultmg: Done with a yaml encoder wrapper that cleans the output.

services_go/observatorium/observatorium.go
thibaultmg: Removed as discussed offline.

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: added

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: fixed

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: fixed

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: I don't really know, just ported what is currently deployed. 
This base64 string means `secret`, it is highly secured then 🔒  . It is mounted in the oauth-proxy and serves as the key for the aes cipher encoding the session cookie set on the client browser.
I am not sure it would work without this option... Do we need to make the secret more secret?
WDYT?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: yup. Removing the list of namespaces here as by default it searches labels in the same namespace as the statefulset. I will remove this namespace also in Observatorium. Because it make things more complex as there, we can modify directly the `DeploymentGenericConfig.Namespace` field. But it wont impact the `DeploymentGenericConfig.Affinity` config that is already generated with the default namespace when we instantiate the object. I don't want to have to synchronise this unnecessarily.

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: I reused what we currently do. This secret is generated in app-interfeace by referencing a vault secret. However, it does not contain credentials in our case as they are injected through env vars using secrets generated by app-sre for s3 resources.
You are then suggesting that we move the bucket configuration secret from vault to here?

philipgough: I'm suggesting that we should consider it for sure. It feels better to me but it is up for discussion within the team

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
philipgough: I see. Makes sense. I don't think we actually need it then. We can generate a secret value at deploy time from OpenShift template. This would negate the need to have the secret at all and reduce the size and complexity of the template IMHO. Check out https://docs.openshift.com/container-platform/4.13/openshift_images/using-templates.html#templates-writing-parameters_using-templates

thibaultmg: Cool feature! Done.

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: This can now be done using the preManifestHooks. This should provide more flexibility as we won't have to handle all these options mapping individually.

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: same as above

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: I have studied this solution. Current implementation looks preferable to me. As I understand, all is done by app-sre that directly pushes the secret in vault and makes it available as a secret. 
If we want to put it in the config, we would still have to do that first, then read the vault data and copy paste into rhobs config.

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
thibaultmg: done

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
thibaultmg: done

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
philipgough: there is direct support for an s3 Secret but not for Thanos object store configuration and others etc. To me at least it is not preferable because it is too ambiguous. If we have a template for the Secret then we know in advance what fields the Secret needs and the values are just injected in. It is a separation of concerns between configuration and secrets.

From an upstream perspective this makes no little sense because now we have a hardcoded Secret values in a template that have no meaning. Their contents is undefined.

That is a personal opinion though and I am happy to. go with consensus. I am interested to hear what others think :)

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,299 @@
...
+          - -openshift-service-account=observatorium-thanos-compact
+          - '-openshift-sar={"resource": "namespaces", "verb": "get", "name": "rhobs",
+            "namespace": "rhobs"}'
+          - '-openshift-delegate-urls={"/": {"resource": "namespaces", "verb": "get",
+            "name": "rhobs", "namespace": "rhobs"}}'
+          - -tls-cert=/etc/tls/private/tls.crt
+          - -tls-key=/etc/tls/private/tls.key
+          - -client-secret-file=/var/run/secrets/kubernetes.io/serviceaccount/token
+          - -cookie-secret=${OAUTH_PROXY_COOKIE_SECRET}
+          - -cookie-secret-file=/etc/proxy/secrets/session_secret
```
philipgough, line 220: We should be able to drop this flag now in favour of the generated value?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,299 @@
...
+              memory: 200Mi
+            requests:
+              cpu: 100m
+              memory: 100Mi
+          terminationMessagePolicy: FallbackToLogsOnError
+          volumeMounts:
+          - mountPath: /etc/tls/private
+            name: compact-tls
+            readOnly: true
+          - mountPath: /etc/proxy/secrets
```
philipgough, line 241: we can drop this?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
```diff
@@ -0,0 +1,299 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  creationTimestamp: null
+  name: observatorium-metrics-compact
+objects:
+- apiVersion: v1
+  data:
+    session_secret: c2VjcmV0
+  kind: Secret
```
philipgough, line 10: IIUC we no longer need this?

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
philipgough: my concern would be incident management. I wouldn't want to have to go through an additional PR & MR CI cycle just to change log level during an emergency 🤔

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: oups 😬 . Deleted

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: done

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: hum that's right, missed the s3 -> thanos object store config step. I agree we must improve the current state

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-store-template.yaml
thibaultmg: changed

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: Missed your reply. Yes I agree... Will add log_level and replicas as openshift params

resources/services/app-sre-stage-01/rhobs/observatorium-metrics-compact-template.yaml
thibaultmg: closing as changed and discussed offline. We will probably have to change this later on.

go.mod
```diff
@@ -2,15 +2,14 @@ module github.com/rhobs/configuration
 
 go 1.21
 
-toolchain go1.21.1
-
```
douglascamata, line 4: Curious question: why is this gone? 🤔

go.mod
```diff
@@ -84,4 +82,5 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace github.com/observatorium/observatorium => github.com/thibaultmg/observatorium v0.0.0-20230922090626-0f7a82fd0910
+// Delete when https://github.com/observatorium/observatorium/pull/543 is merged to main branch
+replace github.com/observatorium/observatorium => github.com/thibaultmg/observatorium v0.0.0-20231002130052-3e0c7849c81f
```
douglascamata, line 86: Let's not forget this TODO. `replace` is evil. 😈

douglascamata (approved):
LGTM :shipit:

go.mod
thibaultmg: I had issues with the CI that was breaking here because I missed to upgrade one of the two Go images used by it. It was yelling at me on this line, I just removed it. At this time, I did not know about the second image 🤷 .
I did not put it back because not needed.

go.mod
thibaultmg: Yup. But it should hit back soon enough so that we don't forget about it. 👊

moadz (commented):
Looks good :) Just a few changes we ned to look at.

PR #603, opened 2023-09-21
Add obsctl reloader alerting rules

douglascamata:
Beware: I apparently cannot type `obsctl-reloader` and end up typing `obsctl-realoder` quite often.


resources/observability/prometheusrules/observatorium-obsctl-reloader-production.prometheusrules.yaml
```diff
@@ -0,0 +1,67 @@
...
+      expr: |
+        (
+          sum_over_time(obsctl_reloader_prom_rule_set_failures_total{reason!="rules_store_error", job="rules-obsctl-reloader"}[5m])
+        /
+          sum_over_time(obsctl_reloader_prom_rule_set_total{job="rules-obsctl-reloader"}[5m])
+        ) or vector(0)
+        > 0.10
+      for: 10m
+      labels:
+        service: telemeter
```
philipgough, line 48: is this correct? I thought it would be for the observatorium-mst service?

philipgough (approved):
lgtm

resources/observability/prometheusrules/observatorium-obsctl-reloader-production.prometheusrules.yaml
douglascamata: Apparently everything we have deploys with `service: telemeter`. There's no such `service: observatorium-mst`. 😬 

It is definitely wrong and we need to fix this at some point.

philipgough: ack, i expected your response so it was marked as lgtm regardless of it. We do have two distinct services onboarded within app-interface though, so yeah, we should probably fix that

PR #605, opened 2023-09-25
metamonitoring: Add scheme to s3 endpoint as required by Tempo operator

philipgough:
See https://github.com/grafana/tempo-operator/blob/a701112a9c1b3557b4c3b3d828c84629b33ca597/apis/tempo/v1alpha1/storage.go#L92-L114

### DIFF
--- a/resources/services/meta-monitoring/tracing-template.yaml
+++ b/resources/services/meta-monitoring/tracing-template.yaml
@@ -16,6 +16,8 @@ parameters:
   - name: S3_BUCKET_NAME
   - name: S3_BUCKET_ENDPOINT
     value: s3.us-east-1.amazonaws.com
+  - name: S3_BUCKET_ENDPOINT_SCHEME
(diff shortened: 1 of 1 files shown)


resources/services/meta-monitoring/tracing-template.yaml
```diff
@@ -39,7 +41,7 @@ objects:
       access_key_id: ${S3_ACCESS_KEY_ID}
       access_key_secret: ${S3_SECRET_ACCESS_KEY}
       bucket: ${S3_BUCKET_NAME}
-      endpoint: ${S3_BUCKET_ENDPOINT}
+      endpoint: '${S3_BUCKET_ENDPOINT_SCHEME}${S3_BUCKET_ENDPOINT}'
```
frzifus, line 44: Is that not part of the `S3_BUCKET_ENDPOINT` value? :x:

PR #606, opened 2023-09-25
metamonitoring: Patch routes for tracing

philipgough:


frzifus (approved):
LGTM

frzifus (commented):
Depending on the running version we have to set the hostname too.

PR #617, opened 2023-10-05
meta-monitoring: enable batchprocessor

frzifus:


resources/services/meta-monitoring/tracing-template.yaml
```diff
@@ -253,5 +255,5 @@ objects:
             pipelines:
               traces:
                 receivers: [otlp, jaeger]
-                processors: [memory_limiter, k8sattributes, resourcedetection]
+                processors: [memory_limiter, k8sattributes, resourcedetection, batch]
```
philipgough, line 258: out of interest, does the order here effect anything?

philipgough (approved):
lgtm

resources/services/meta-monitoring/tracing-template.yaml
frzifus: I assume not with the currently used processors - but in general yes. 

e.g.:

> This processor must be placed in pipelines after any processors that rely on context, e.g. k8sattributes. It reassembles spans into new batches, causing them to lose their original context.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3f5d64cd86a9ba9c9c70d14c50aa9d0ef766efc1/processor/tailsamplingprocessor/README.md

PR #619, opened 2023-10-09
Add rhelemeter namespace to the Telemeter dashboard

douglascamata:
Most of the Telemeter dashboard can be reused for Rhelemeter, as both support the remote write handler. Any further improvements to this reusability can/may be done down the line. 

Upgrades to the Observatorium docs, logging CRDs and rules were necessary to pass the build.


thibaultmg (approved):
LGTM

PR #626, opened 2023-10-17
Remove `exhaustion` label from SLO rules

douglascamata:
This is added by new versions of Pyrra and it's not in the schema accepted by app-sre, which makes the build fail over there.


douglascamata:
Closing at the field in now allowed in the app-sre prometheus rule schema.

PR #638, opened 2023-11-07
[rhobs]: fix empty probes regression, fix store init container image

thibaultmg:


resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-store-default-template.yaml
```diff
@@ -781,7 +777,7 @@ objects:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
-          image: quay.io/app-sre/ubi8-ubi
+          image: registry.redhat.io/openshift4/ose-cli:v4.12
```
philipgough, line 780: I think you might have difficulty with deploying this via app-interface unless it is already allowed to pull through this registry.

I don't want to derail this but I wonder at the offset are we better off to do a like for like deployment and drop the dynamic hashmod configuration until we are comfortably rolled out?

Thoughts @douglascamata ?

resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-store-default-template.yaml
```diff
@@ -259,13 +257,11 @@ objects:
           - --verbose=true
           image: quay.io/app-sre/memcached:1.5
           imagePullPolicy: IfNotPresent
-          livenessProbe: {}
           name: observatorium-thanos-store-index-cache-memcached-default
           ports:
           - containerPort: 11211
             name: client
             protocol: TCP
-          readinessProbe: {}
```
philipgough, line 268: do we not need these probes to be configured?

resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-store-default-template.yaml
thibaultmg: It's 90% ready... Now it is easier to go with that

resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-store-default-template.yaml
thibaultmg: When it is empty like that, app-sre cries. If there is none, it works 🤷‍♂️

resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-store-default-template.yaml
thibaultmg: if this iage causes problems, I can still revert to the way you were doing it @philipgough , without calling oc.

resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-store-default-template.yaml
philipgough: Sorry, I mostly mean that we should probably keep the existing probes that are in place like https://github.com/rhobs/configuration/blob/main/resources/services/observatorium-metrics-template.yaml#L2887-L2906

resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-store-default-template.yaml
thibaultmg: Can also use quay.io/openshift/origin-cli:4.15 and add this to the imagePatterns of the saas file. Wdyt?

douglascamata: @philipgough I think it's a good time to experiment with the image from Red Hat's Registry, as there's no immediate urgency in this and there are many more steps to be completed before we can use everything. This is an official OpenShift image and I see no good reason for app-interface to forbid us from pulling it. 

I think the dynamic sharding is very important for rollouts or any other situation where one Store GW shard/replica goes kaput -- we need to be able to keep serving all queries that would otherwise go to the missing shard.

If getting this to deploy and run takes too long, then I think we can fallback to the "static" version of the solution.

resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-store-default-template.yaml
thibaultmg: This is only removing empty probes. Here memcached does not have any. Even in our current deployment (unless I missed something 😲 ). Thanos containers still have their probes

philipgough: No, sorry my mistake. All good then

philipgough (approved):

PR #641, opened 2023-11-08
[rhobs]: add quey pull secrets 

thibaultmg:
Adds quay pull secret to all components lock in imagePullBackOff state. They are all pulling images from `quay.io/app-sre`


douglascamata (approved):
lgtm

PR #645, opened 2023-11-08
[rhobs]: fix hashring name is base cm

thibaultmg:
The cli option in the controller sets both input and output name in the CM. Fixing the input...


douglascamata (approved):
🦖

PR #653, opened 2023-11-24
[rhobs]: Add rhobs query

thibaultmg:
Adds thanos query deployment to rhobs


resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-query-rule-default-template.yaml
```diff
@@ -0,0 +1,318 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  creationTimestamp: null
+  name: observatorium-thanos-query-rule-default
+objects:
+- apiVersion: route.openshift.io/v1
+  kind: Route
```
philipgough, line 8: I am trying to understand why we expose this directly via the internet? Is it for us (rhobs) to access the Thanos UI?

resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-query-rule-default-template.yaml
```diff
@@ -0,0 +1,318 @@
...
+- apiVersion: apps/v1
+  kind: Deployment
+  metadata:
+    creationTimestamp: null
+    labels:
+      app.kubernetes.io/component: query-layer
+      app.kubernetes.io/instance: observatorium
+      app.kubernetes.io/name: thanos-query-rule
+      app.kubernetes.io/part-of: observatorium
+      app.kubernetes.io/version: v0.32.4
```
philipgough, line 123: i guess this should be v0.32.5?

resources/services/app-sre-stage-01/rhobs/default/observatorium-metrics-query-rule-default-template.yaml
thibaultmg: Yes this is what I was thinking, for debugging purpose

PR #661, opened 2023-12-14
Add insights-qa ssh key secret to host-metering e2e test template

tpapaioa:
Add the insights-qa ssh key to the host-metering e2e test template. The test container needs this key in order to connect to cloud VMs that are provisioned during the test run.


philipgough (approved):
Lgtm

PR #665, opened 2023-12-21
[rhobs/stage]: add missing volume types for statefulSets

thibaultmg:
AlertManager and ruler lack volume types. Adding them.


jacobbaungard (approved):
:+1:

PR #669, opened 2024-01-05
[rhobs/stage]: fix tenants secret

thibaultmg:
Embeds several small fixes:

- Adds the secret needed by the OPA AMS sidecar of observatorium-api
- Fixes the default port of the query frontend. Currently makes the pod crashlooping as the readiness probe does not work.
- Fixes the serviceName of the alertmanager statefulset to the cluster service. Currently, replicas cannot communicate for HA setup.

Many fixes are in Observatorium and brought here by updating the commit hash in the replace directive of the go.mod. This Observatorium PR is not ready yet and will be pushed once everything is deployed on our side, validating the configurations generated with it.


resources/services/app-sre-stage-01/rhobs/observatorium-api-template.yaml
```diff
@@ -1502,46 +1518,57 @@ objects:
...
+    namespace: rhobs
+  stringData:
     config.yaml: |
       tenants:
       - name: appsre
         id: 3833951d-bede-4a53-85e5-f73f4913973f
         oidc:
-          clientID: ${CLIENT_ID}
-          clientSecret: ${CLIENT_SECRET}
+          clientID: ${AMS_OIDC_CLIENT_ID}
```
philipgough, line 1538: is this correct? aside from telemeter tenant there should be no other tenant using AMS to authn/authz - those will be issued against RH SSO

thibaultmg: ok, I struggled to understand why these params are there. Let's discuss it offline

thibaultmg: Changed the tenant oidc parameter as discussed. Let me know if I need to change more 🙂

philipgough (approved):
lgtm

PR #673, opened 2024-01-15
[receive]: allow configuration of too-far-in-the-future

douglascamata:
The default value is `0s`, which leaves the feature disabled.


philipgough (approved):
lgtm

PR #678, opened 2024-02-05
RHOBS-961 decreasing memory due to node limit

moadz:
We're running on m5.2xlarge that only have 32Gi of memory. Fixing so the pods can be scheduled.


saswatamcode (approved):
lgtm, minus conflicts

moadz:
Conflicts fixed!

PR #679, opened 2024-02-06
Min ready seconds inc

philipgough:


moadz:
Just so we don't forget this later, are you happy to do this for receive-ingestors as well on the rhobs namespace go-configs?

PR #683, opened 2024-03-15
Decomission Loki tenants

philipgough:
This change removes logging/tracing flags and params from the observatorium template and regenerates the rbac and tenant config


douglascamata (approved):
LGTM 🪦

PR #688, opened 2024-04-24
MON-3707: Adding ipsec state metric into telemetry

JoshSalomon:
Just the last part of adding this metric to Telemtery.
The metric was added to CNO in https://github.com/openshift/cluster-network-operator/pull/2346
and to the monitoring operator in https://github.com/openshift/cluster-monitoring-operator/pull/2326


philipgough (approved):
/lgtm

PR #694, opened 2024-07-10
Adding store-gw and memcached configuration to match existing datahub store path

moadz:
We are offsetting these store-gw's by 14d as Telemeter's existing hot path will fulfill queries for that duration until we decommission those components.

### DIFF
--- a/resources/services/telemeter-prod-01/rhobs/telemeter/observatorium-metrics-store-telemeter-template.yaml
+++ b/resources/services/telemeter-prod-01/rhobs/telemeter/observatorium-metrics-store-telemeter-template.yaml
@@ -83,7 +83,7 @@ objects:
     name: observatorium-thanos-store-bucket-cache-memcached-telemeter
     namespace: rhobs
   spec:
-    replicas: 1
+    replicas: 5
     selector:
       matchLabels:
         app.kubernetes.io/component: store-bucket-cache
@@ -135,10 +135,10 @@ objects:
             protocol: TCP
           resources:
             limits:
-              memory: 3Gi
+              memory: 12Gi
             requests:
-              cpu: 500m
-              memory: 2Gi
+              cpu: "6"
+              memory: 10Gi
           terminationMessagePolicy: FallbackToLogsOnError
         - args:
           - --memcached.address=localhost:0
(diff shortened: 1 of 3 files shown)


resources/services/telemeter-prod-01/rhobs/telemeter/observatorium-metrics-store-telemeter-template.yaml
```diff
@@ -292,10 +293,11 @@ objects:
             protocol: TCP
           resources:
             limits:
-              memory: 3Gi
+              cpu: "7"
```
thibaultmg, line 296: I thought it was recommended to not set limits?

resources/services/telemeter-prod-01/rhobs/telemeter/observatorium-metrics-store-telemeter-template.yaml
```diff
@@ -292,10 +293,11 @@ objects:
...
+              memory: 12Gi
             requests:
-              cpu: 500m
-              memory: 2Gi
+              cpu: "6"
```
thibaultmg, line 299: It's a huge increase isn't it? It's 6*5 cpus for each cache (store and index).
Given the type of worker in this cluster (r5.2xlarge with 8 vCPUs), it's one worker per replicas, so 10 workers for the cache.
If it's what is needed for DH it's good for me, just checking?

moadz: Yeah it seems like an overestimation but it's going to be fulfilling queries for 4 years worth of Telemetry (telemeter just does 14d at the moment). I'm overestimating just so that index caches can be built. 

Eventually i want to get rid of the observatorium-mst store path, so we should recycle some of those pods as well. In the interim it will be a bit unwieldy.

resources/services/telemeter-prod-01/rhobs/telemeter/observatorium-metrics-store-telemeter-template.yaml
moadz: You're correct! Fixed.

thibaultmg (approved):
/lgtm

PR #696, opened 2024-08-15
Adding store.enable-index-header-lazy-reader to store-gw template.

moadz:
Reference: https://thanos.io/tip/components/store.md/


philipgough (approved):
/lgtm

PR #700, opened 2024-10-16
Update Telemeter metrics allow-list

simonpasquier:


configuration/telemeter/metrics.json
```diff
@@ -1,6 +1,5 @@
 [
-  "{__name__=\":apiserver_v1_image_imports:sum\"}",
```
simonpasquier, line 2: I'm surprised that we're removing so many old metrics. Let me investigate.

simonpasquier: @philipgough

philipgough (approved):
/lgtm

saswatamcode (approved):

PR #713, opened 2024-12-25
Removing matcher param to get recording rule blocks and adding compaction level max to avoid copying non-compacted blocks (for-now).

moadz:
* Removing matcher param to make sure we pick up the non-tenanted recording rule blocks
* Setting min-compaction to 3 to avoid copying blocks that later have to be re-compacted/duplicated over time (e.g. we copy the raw blocks from receivers, in two hours we also copy the compacted blocks by telemeter compaction and we have to double-compact)

After this - I'll raise a PR to repoint to the telemeter-lts bucket (>120Gi of DH metrics), then I will repoint the rhobs namespace thanos compactor to that bucket as well so we can monitor compaction.


philipgough (approved):
/lgtm

PR #716, opened 2025-01-14
Mage crds

philipgough:
This change adds a dep on https://magefile.org/ which is useful for generating specific components as we plan to structure the templates in layers.

This includes the build for the upstream CRDs which are bundled into the https://github.com/thanos-community/thanos-operator repo.

For now, we can sync from `main` for which the output here for `mage crds` is committed.


saswatamcode (approved):
Nice!

saswatamcode (approved):

PR #717, opened 2025-01-14
Mage alertmanager

philipgough:
This change adds mage target to generate alertmanager configuration
This includes the file generated as `mage stage:alertmanager` as per the second commit


saswatamcode (approved):
Lgtm minus some of the build fails

PR #733, opened 2025-02-11
mage: memcached for api gateway

philipgough:


magefiles/cache.go
```diff
@@ -0,0 +1,301 @@
...
+			Kind:       "StatefulSet",
+			APIVersion: "apps/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      config.Name,
+			Namespace: config.Namespace,
+			Labels:    labels,
+		},
+		Spec: appsv1.StatefulSetSpec{
+			Replicas: &config.Replicas,
```
saswatamcode, line 220: Could you work the image/rep/resource fields into templates.go? Would be easier to have unified place for doing things

philipgough: i can. i wonder do you think it is polluting the main things we need to change though, ie thanos? or you prefer the consistency?

saswatamcode: We can have sub-structs for the thanos ones. But yeah prefer to have things be consistent and in one place, for quick, small changes in future 🙂

saswatamcode (approved):

PR #743, opened 2025-02-14
Temporarily remove stage CRs

saswatamcode:
This blocks cluster upgrades as we aren't operational on it yet


saswatamcode:
This needs app-interface

PR #745, opened 2025-02-17
MON-4151: Update Telemeter metrics allow-list

falox:


resources/services/telemeter-template.yaml
```diff
@@ -126,6 +126,7 @@ objects:
           - --whitelist={__name__="cluster:eo_es_data_nodes_total:max"}
           - --whitelist={__name__="cluster:eo_es_documents_created_total:sum"}
           - --whitelist={__name__="cluster:eo_es_documents_deleted_total:sum"}
+          - --whitelist={__name__="cluster:cluster:health:group_severity:count"}
```
juzhao, line 129:
```suggestion
          - --whitelist={__name__="cluster:health:group_severity:count"}
```

falox: thanks @juzhao :man_facepalming:

PR #824, opened 2025-03-28
Add MAX_AGE to rclone job, default to 4w. Will only replicate objects…

moadz:
… younger than 4w.

https://rclone.org/commands/rclone_copy/

     ` --max-age Duration                    Only transfer files younger than this in s or suffix ms|s|m|h|d|w|M|y (default off)`


philipgough (approved):
/lgtm

PR #850, opened 2025-04-16
Sync telemeter allowlist

philipgough:


juzhao:
@philipgough 
below metrics are in server allow list, they are with different types
```
type:tempo_operator_tempostack_multi_tenancy:sum
type:opentelemetry_collector_receivers:sum
type:opentelemetry_collector_exporters:sum
type:opentelemetry_collector_processors:sum
type:opentelemetry_collector_extensions:sum
type:opentelemetry_collector_connectors:sum
```
example
```
          - --whitelist={__name__="type:opentelemetry_collector_connectors:sum",type=~"spanmetricsconnector|forwardconnector"}
          - --whitelist={__name__="type:opentelemetry_collector_connectors:sum",type=~"spanmetrics|forward"}
 ```
we should leave the second one based on PR
https://github.com/openshift/cluster-monitoring-operator/pull/2583/files
the first one was the original setting, the second one is the PR has been updated, I think we should only keep the second one, but keep them both is also fine, not sure if we have standard for such issue?

PR #932, opened 2025-06-09
RHOBS-904: Standalone Store and Compactor for RH Resource Optimisation (ROS) Managed Service

moadz:
The reason we only need these two components is: 
* They will upload their blocks asynchronously via ROS managed service
* They will use RHOBS as a read-only tenant to access those blocks

Next: 
* Generate read only credentials
* Generate shared bucket credentials for the tenant

### DIFF
--- a/resources/services/rhobs-thanos-operator/staging/rhobs.yaml
+++ b/resources/services/rhobs-thanos-operator/staging/rhobs.yaml
@@ -78,7 +78,6 @@ objects:
       lazyReaderIdleTimeout: 5m
     logFormat: logfmt
     logLevel: info
-    maxTime: -22h
     objectStorageConfig:
       key: thanos.yaml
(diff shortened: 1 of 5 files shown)


resources/services/rhobs-thanos-operator/staging/rhobs.yaml
```diff
@@ -705,6 +890,9 @@ objects:
     storeLimitsOptions:
       storeLimitsRequestSamples: 627040000
       storeLimitsRequestSeries: 1000000
+    timeRangeConfig:
+      maxTime: -2w
+      minTime: -90d
     version: 03c9fefbda1b33830a950bd28484fa0a1e039555
```
moadz, line 896: @philipgough i'm assuming this is the CR changes you made and this update is fine? Do i need to bump operator refs?

philipgough (approved):
/lgtm

PR #942, opened 2025-07-29
Adding Cluster Matrix for RHOBS Configuration Generation

moadz:
Details in README.md


magefiles/clusters.go
```diff
@@ -0,0 +1,119 @@
...
+		return fmt.Errorf("cluster must have at least one build step")
+	}
+	return nil
+}
+
+// DefaultBuildSteps returns the standard build pipeline
+func DefaultBuildSteps() []string {
+	return []string{
+		StepThanos,
+		StepCRDS,
```
philipgough, line 74: I think we should probably disambiguate the steps between signals somehow, like metrics and logs and a cluster would contain steps per signal

magefiles/cluster_app_sre_stage_01.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+const (
+	ClusterAppSREStage01 ClusterName = "app-sre-stage-01"
```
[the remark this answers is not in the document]
moadz: This is definitely the goal. If we're happy with the rest of the PR, i'll split it out to it's own package. 

I want to move the template/config utils out as well. Didn't want to make the diff too complicated.

magefiles/servicemonitors.go
```diff
@@ -11,6 +11,21 @@ import (
 	"k8s.io/utils/ptr"
 )
 
+func (b Build) ServiceMonitors(config ClusterConfig) {
```
philipgough, line 14: the operator generates the service monitors, so its a shame we have to do this. I will raise it with app-sre about the possibility of the prometheus scraping servicemonitors in other namespaces

philipgough (changes requested):
Just some alternatives here to the mage commands I am suggesting

# Build all registered clusters
mage build:allClusters -> could just be mage build:cluster and if no name is passed, the default would be all clusters

# Build specific cluster by name
mage build:buildCluster my-cluster-name -> could just be mage build:cluster my-cluster-name

# Build all clusters in an environment
mage build:buildEnvironment staging -> mage build:environment staging
mage build:buildEnvironment production  -> mage build:environment production
mage build:buildEnvironment integration


mage build:listAvailableSteps -> build makes little sense here its not building anything. Suggest using mage list:steps or mage listSteps/listavailableSteps

# Show build steps for each cluster
mage build:listClusterSteps -> as above, nothing is being built here so we dont need to namespace it.

mage local:build  -> this is legacy but we could just drop the local stuff and have a static local template to play with. Unless it ties into the other stuff that is being used and generated it just becomes redundant

magefiles/servicemonitors.go
moadz: Yeah I agree :/

magefiles/magefile.go
```diff
@@ -12,13 +13,141 @@ type (
...
+		err := b.Gateway(cfg)
+		if err != nil {
+			return err
+		}
+		return nil
+	},
+}
+
+// ExecuteSteps executes a list of build steps for a cluster
+func (b Build) ExecuteSteps(steps []string, cfg ClusterConfig) error {
```
philipgough, line 71: I would avoid exporting this function so it doesnt show up on `mage -l`

magefiles/servicemonitors.go
saswatamcode: Yeah this is a pain

magefiles/template.go
```diff
@@ -62,6 +155,162 @@ const (
...
+			"STORE_DEFAULT":            thanosImage,
+			"RECEIVE_ROUTER":           thanosImage,
+			"RECEIVE_INGESTOR_DEFAULT": thanosImage,
+			"RULER":                    thanosImage,
+			"COMPACT_DEFAULT":          thanosImage,
+			"QUERY":                    thanosImage,
+			"QUERY_FRONTEND":           thanosImage,
+			jaeger:                     "registry.redhat.io/rhosdt/jaeger-agent-rhel8:1.57.0-10",
+			"api_cache":                "docker.io/memcached:1.6.17-alpine",
+			"memcached_exporter":       memcachedExporterImage,
```
saswatamcode, line 173: Maybe they should be capitalized?

Also api cache image can just be memcached image right?

magefiles/template.go
```diff
@@ -62,6 +155,162 @@ const (
...
+		},
+		Versions: ParamMap[string]{
+			"STORE_DEFAULT":            thanosVersionProd,
+			"RECEIVE_ROUTER":           thanosVersionProd,
+			"RECEIVE_INGESTOR_DEFAULT": thanosVersionProd,
+			"RULER":                    thanosVersionProd,
+			"COMPACT_DEFAULT":          thanosVersionProd,
+			"QUERY":                    thanosVersionProd,
+			"QUERY_FRONTEND":           thanosVersionProd,
+			apiCache:                   memcachedTag,
```
saswatamcode, line 183: Let's keep the keys consistent between the maps

magefiles/cluster_rhobsi01uw2.go
```diff
@@ -0,0 +1,22 @@
...
+		AMSUrl:      "https://api.openshift.com",
+		Templates:   rhobsi01uw2TemplateMaps(),
+		BuildSteps:  DefaultBuildSteps(),
+	})
+}
+
+// rhobsi01uw2TemplateMaps returns template mappings specific to the rhobsi01uw2 integration cluster
+func rhobsi01uw2TemplateMaps() TemplateMaps {
+	// Start with integration base template and override only what's different
+	return DefaultBaseTemplate()
```
saswatamcode, line 21: Don't we need to override the object storage here?

magefiles/magefile.go
```diff
@@ -12,13 +13,141 @@ type (
...
+		if err := b.ExecuteSteps(cfg.BuildSteps, cfg); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// ListAvailableSteps lists all available build steps
+func (b Build) List() {
+	fmt.Println("Available build steps:")
```
saswatamcode, line 130: We can remove the faillint rule here or use Fprintf with os.Stdout

magefiles/clusters.go
saswatamcode, line 1: Since we have the clusters nomenclature now, maybe we can do away with the "local" stuff I added last time?

saswatamcode (approved):
This looks amazing! Thanks for the hard work! 💪🏻 

LGTM, mod some comments + @philipgough's comments above.

PR #959, opened 2025-08-08
Add cache secrets for memcached and use config structs from Thanos

saswatamcode:


magefiles/secrets.go
```diff
@@ -115,10 +124,19 @@ func secrets(gen *mimic.Generator, ns string) {
 
 func (b Build) Secrets(config clusters.ClusterConfig) {
 	gen := b.generator(config, "secrets")
+	cacheSecrets := memcachedCacheSecrets(config.Namespace)
+
+	secrets := []runtime.Object{
+		thanosObjectStoreSecret("observatorium-mst-thanos-objectstorage", config.Namespace),
```
moadz, line 130: nit: Should we rename this secret? It's a redundant name.

moadz (approved):
LGTM. Should we try and get elasticache configured via app-interface?

